PR #29, opened 2026-04-26
fix: load user agent types from dashboard config

### LINKED ISSUE #28: Custom agent types in dashboard config.toml are silently ignored
## Problem

When users add `[agents.<name>]` sections to `~/.config/lince-dashboard/config.toml` (e.g. `[agents.zai]`, `[agents.codex]`), the dashboard wizard does not show them. Only the built-in `claude` type from `agents-defaults.toml` is available.

The wizard's agent type selector is only shown when `available_agent_types.len() > 1`, so with only `claude` loaded the selector step is silently skipped and users see a hardcoded `Type: claude` on the confirm screen with no way to change it.

## Root Cause

`load_agent_defaults_async()` in `config.rs` reads two files separated by NUL:
1. `~/.config/lince-dashboard/agents-defaults.toml` (built-in defaults)
2. `~/.agent-sandbox/config.toml` (sandbox config)

The second file is the **sandbox** configuration, which contains bwrap jail settings (`home_ro_dirs`, `persist_toolchains`, etc.) — not dashboard agent type definitions. User-defined agent types like `[agents.zai]` are placed in the **dashboard** config (`~/.config/lince-dashboard/config.toml`), which is never read during agent type loading.

## Expected Behavior

`[agents.*]` sections in `~/.config/lince-dashboard/config.toml` should be merged with the defaults, making custom agent types available in the wizard.

## Reproduction

1. Add `[agents.zai]` section to `~/.config/lince-dashboard/config.toml`
2. Run `zd`, press `N` for full wizard
3. Observe: no agent type selector — only `claude` is available

paoloantinori:
## Summary
- Fix `load_agent_defaults_async()` to read `~/.config/lince-dashboard/config.toml` (dashboard config) for user `[agents.*]` overrides instead of `~/.agent-sandbox/config.toml` (sandbox config)
- Add example `[agents.zai]` / `[agents.zai-unsandboxed]` entries to the source `config.toml`

## Test plan
- [ ] Add `[agents.zai]` section to `~/.config/lince-dashboard/config.toml`
- [ ] Build WASM: `cd lince-dashboard/plugin && PATH="$HOME/.cargo/bin:$PATH" $HOME/.cargo/bin/cargo build --target wasm32-wasip1 --release`
- [ ] Install: `cp target/wasm32-wasip1/release/lince-dashboard.wasm ~/.config/zellij/plugins/`
- [ ] Run `zd`, press `N` for wizard
- [ ] Verify agent type selector shows Zai alongside Claude Code

Closes #28

### DIFF
--- a/lince-dashboard/config.toml
+++ b/lince-dashboard/config.toml
@@ -64,3 +64,27 @@ sandbox_backend = "auto"
 # bwrap_conflict = false
 # disable_inner_sandbox_args = []
 
+# Zai (cc-mirror) agent types
+[agents.zai]
+command = ["agent-sandbox", "run", "-p", "{project_dir}", "--id", "{agent_id}"]
+pane_title_pattern = "agent-sandbox"
+status_pipe_name = "claude-status"
+display_name = "Zai"
+short_label = "ZAI"
+color = "blue"
+sandboxed = true
+has_native_hooks = true
+home_ro_dirs = ["~/.cc-mirror/zai/config/"]
+profiles = ["__discover__"]
+
+[agents.zai-unsandboxed]
+command = ["zai"]
+pane_title_pattern = "zai"
+status_pipe_name = "claude-status"
+display_name = "Zai (unsandboxed)"
+short_label = "ZAU"
+color = "blue"
+sandboxed = false
+has_native_hooks = true
+profiles = ["__discover__"]
+
--- a/lince-dashboard/config.toml
+++ b/lince-dashboard/config.toml
@@ -63,28 +63,3 @@ sandbox_backend = "auto"
 # home_rw_dirs = []
 # bwrap_conflict = false
 # disable_inner_sandbox_args = []
-
-# Zai (cc-mirror) agent types
-[agents.zai]
-command = ["agent-sandbox", "run", "-p", "{project_dir}", "--id", "{agent_id}"]
-pane_title_pattern = "agent-sandbox"
-status_pipe_name = "claude-status"
-display_name = "Zai"
-short_label = "ZAI"
-color = "blue"
-sandboxed = true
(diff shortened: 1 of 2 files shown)


lince-dashboard/config.toml
```diff
@@ -64,3 +64,27 @@ sandbox_backend = "auto"
 # bwrap_conflict = false
 # disable_inner_sandbox_args = []
 
+# Zai (cc-mirror) agent types
```
maeste: This should go in your own configuration, not in the repository. We already have an example for custo agent and I think it's sufficient for the repo file

lince-dashboard/config.toml
```diff
@@ -64,3 +64,27 @@ sandbox_backend = "auto"
...
+status_pipe_name = "claude-status"
+display_name = "Zai"
+short_label = "ZAI"
+color = "blue"
+sandboxed = true
+has_native_hooks = true
+home_ro_dirs = ["~/.cc-mirror/zai/config/"]
+profiles = ["__discover__"]
+
+[agents.zai-unsandboxed]
```
maeste: see above

maeste (changes requested):

paoloantinori:
Thanks for the review! You're right — the `[agents.zai]` and `[agents.zai-unsandboxed]` sections were specific to my setup and don't belong in the repo config.

I've removed them in the latest push. The `config.rs` fix remains — that's the actual bug: `load_agent_defaults_async()` was reading `~/.agent-sandbox/config.toml` for user agent overrides instead of `~/.config/lince-dashboard/config.toml`. With this fix, users can add custom agent types in their own dashboard config and they'll show up in the wizard.

maeste (approved):

Issue #30, opened 2026-04-27
Consider agentsh as sandbox backend


maeste:
https://www.agentsh.org/

This probably make necessary to revisit the backend selection in "N" wizard.
The spi backend should be already ok to support another backend

PR #34, opened 2026-04-28
feat(sandbox): nono backend improvements — Vertex AI, marketplace self-heal, node/npx support

kpiwko:
## Summary

- **Vertex AI credentials**: wire `sandbox_home()`, set `NPM_CONFIG_CACHE`, inject
  `GOOGLE_APPLICATION_CREDENTIALS` from the ADC JSON file (more reliable than
  `CLOUDSDK_CONFIG` alone), add `.config/gcloud` to read-only sandbox paths
- **Marketplace self-heal**: add `_fix_sandbox_known_marketplaces()` called on every
  `nono-sync` — corrects stale `installLocation` paths and converts SSH git sources
  to HTTPS (SSH agent socket is unavailable inside the Seatbelt sandbox)
- **node/npx support**: detect the active node binary across fnm, nvm, Homebrew,
  and system installs; add it to the synthetic PATH and to nono `allow_paths` so
  npx-based MCP servers (e.g. context7) can start inside the sandbox

## Test plan
- [ ] `agent-sandbox nono-sync` runs without errors and writes profiles to `~/.config/nono/profiles/`
- [ ] `agent-sandbox run` with nono backend launches Claude Code inside the sandbox
- [ ] context7 MCP server starts (confirms npx/node on PATH)
- [ ] Vertex AI authentication works (confirms `.config/gcloud` in read paths)
- [ ] `agent-sandbox nono-sync` after moving sandbox config corrects `installLocation` paths in `known_marketplaces.json`

### DIFF
--- a/sandbox/agents-defaults.toml
+++ b/sandbox/agents-defaults.toml
@@ -35,7 +35,10 @@
 command = "claude"
 default_args = ["--dangerously-skip-permissions"]
 env = {}
-home_ro_dirs = []
+# .config/gcloud holds Google Cloud ADC credentials (application_default_credentials.json)
+# needed for Vertex AI authentication.  Users without a Vertex AI setup can override
+# this to [] in their config.toml under [agents.claude].
+home_ro_dirs = [".config/gcloud"]
 home_rw_dirs = []
 bwrap_conflict = false
 disable_inner_sandbox_args = []
--- a/.gitignore
+++ b/.gitignore
@@ -27,3 +27,4 @@ Thumbs.db
 
 #Lince dashboard files
 .lince-dashboard
+
--- a/sandbox/agent-sandbox
+++ b/sandbox/agent-sandbox
@@ -71,13 +71,7 @@ persist_toolchains = true
 # Auto-detect PATH entries under $HOME and expose read-only
 auto_expose_path = true
 
-# Extra home subdirectories to expose read-only (relative to $HOME)
-# .config/gcloud is needed for Vertex AI / Google Cloud authentication
-home_ro_dirs = [".config/gcloud"]
-
 # Extra home subdirectories to expose read-write (relative to $HOME)
-# Use this if read-only access to a directory is insufficient, e.g. if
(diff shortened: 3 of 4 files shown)


sandbox/agents-defaults.toml
```diff
@@ -35,7 +35,10 @@
 command = "claude"
 default_args = ["--dangerously-skip-permissions"]
 env = {}
-home_ro_dirs = []
+# .config/gcloud holds Google Cloud ADC credentials (application_default_credentials.json)
+# needed for Vertex AI authentication.  Users without a Vertex AI setup can override
```
maeste, line 39: I would prefer the opposite, adding them if they are using it. Otherwise, there is a case where they have gcloud credentials accessed from the agent with keys not related to the agent. BTW is necessary to share the whole directory? Any chance to be more fine grained?

maeste (changes requested):

maeste:
Regarding the test plan:
Did you run it?
I think we should add also some tests with bwrap to ensure any changes to agents-default.toml create issues.

Thanks a lot for the PR and the interest

sandbox/agents-defaults.toml
kpiwko: makes sense, let me update that.

kpiwko:
@maeste yes, I have executed the plan however I have found some issues with sandboxing home together with nono - for instance, agent status and tokens were not showing in zillij dashboard (because `jq` and `zellij` were missing in the syntentic `PATH`

I have some code changes that I will add to this PR. That said, I'll probably mark this a draft and I will start exploring using [openshell](https://github.com/NVIDIA/OpenShell/tree/main) as a sandbox instead of `nono` because I'm not sure about scalability of the solution I'm pursuing with SANDBOX_HOME.

maeste:
great idea to explore also openshell. Please create an issue for that so we can track there.
We already have #30 for agentsh as backend. 

BTW on which platform are you? I suppose mac since you are using nono.

Issue #54, opened 2026-05-05
Sandbox-policy fragments: support extends/inheritance between fragments


maeste:
**Status**: design / future work. Filed during execution of #48 (lince-98) when we added the `[security] allow_domains` allowlist to the bwrap credential proxy.

## Context

Today, sandbox-policy fragments under \`sandbox/profiles/<name>.toml\` are flat: when the user passes \`agent-sandbox run --sandbox-level NAME\`, the fragment \`NAME.toml\` is loaded and deep-merged on top of the user's resolved config. There is **no inheritance between fragments** — each fragment is self-contained.

For \"paranoid + a few extra allowed domains\" the workflow today is:

- **Option 1 (recommended)**: keep \`sandbox_level = \"paranoid\"\` and add the extra domains in your personal \`~/.agent-sandbox/config.toml\` under \`[security] allow_domains\` — they get appended via the existing list-merge semantics.
- **Option 2**: write a self-contained custom fragment that duplicates everything in \`paranoid.toml\` plus your extras.

Option 2 is verbose and error-prone (paranoid's defaults can drift).

## Proposal

Add an \`extends = \"<other-fragment-name>\"\` field at the top of a sandbox-policy fragment. When loading \`X.toml\`, if it has \`extends = \"Y\"\`, recursively load \`Y.toml\` first and deep-merge \`X\` on top of it (same semantics as nono profile inheritance).

Concrete example:

\`\`\`toml
# ~/.agent-sandbox/profiles/paranoid-with-pypi.toml
extends = \"paranoid\"

[security]
allow_domains = [\"pypi.org\", \"files.pythonhosted.org\"]
# Append-merged onto paranoid's allow_domains, which already contains
# api.anthropic.com — final list has both.
\`\`\`

Then \`agent-sandbox run --sandbox-level paranoid-with-pypi\` produces a config that's paranoid + the two extra domains.

## Out of scope for this issue

- Changing the list-append set (\`home_ro_dirs\`, \`allow_domains\`, etc.) — already in place, works as expected.
- Adding new fragment-level fields beyond \`extends\` — keep the surface minimal.
- Auto-detecting filename conventions (\`paranoid-XXX\` → extends paranoid). Filename remains arbitrary; \`extends\` is explicit.

## Acceptance criteria

- [ ] \`extends = \"<name>\"\` field at fragment top level, optional, single string
- [ ] Cycle detection (A extends B extends A) → hard error, don't loop
[...]

Issue #53, opened 2026-05-04
Wizard 'N' picker: expose sandbox_level / sandbox_backend selection


maeste:
**Status**: needs design / brainstorming. **Do not start implementation yet.**

Related: #47 (umbrella), #48 (claude prototype), #49 #50 #51 #52 (per-agent follow-ups).

## Context

After the three-level sandbox refactor (#47 + children), each coding agent has a **single** `[agents.X]` entry in `agents-defaults.toml` plus two attributes:

- `sandbox_level = "paranoid" | "normal" | "permissive"`
- `sandbox_backend = "bwrap" | "nono"`

The current opt-in mechanism for non-default levels is **commented templates** in the TOML — users uncomment what they want. This works but feels clunky compared to a wizard step.

With one entry per agent, the **N-picker wizard** could realistically expose level and backend as interactive choices at spawn time, instead of requiring users to edit the TOML.

## Open design questions (for brainstorming)

1. **When to ask**: every spawn, or only on first spawn of an agent in the session? Remembered per-project?
2. **What to show**: a simple level picker (3 radio choices), level + backend on separate steps, or one combined choice (`paranoid-bwrap`, `permissive-nono`, ...)? The latter scales 3 × N_backends and gets noisy.
3. **Backend default**: keep auto-by-OS hidden unless user expands an "advanced" section, vs. always visible?
4. **Discoverability of paranoid**: do we want to *encourage* paranoid (security-first default for new users), or keep `normal` as the picker default and let users opt up?
5. **Rendering inside Zellij plugin**: the picker today is a small list. Adding a multi-step flow (agent → level → backend) needs UX design — popups, inline expand, preview of what each level grants/denies?
6. **Bypass for power users**: if `agents-defaults.toml` already pins `sandbox_level` for an agent, do we skip the wizard or still allow override per-spawn?
7. **Persistence**: if user picks paranoid for a session, is the choice scoped to the agent instance (each spawn re-asks) or persisted in `~/.config/lince-dashboard/state.toml`?
8. **Visual hint of current level**: in the dashboard, show a badge/icon next to a running agent indicating its sandbox level (so the user knows what they spawned). Same as #2 but on the running side.

## Out of scope

- Implementation. This issue is for design discussion. A separate issue/task tracks the implementation once the design lands.
- The TOML side (`agents-defaults.toml` schema) — already covered by #47/#48.

## When to revisit

[...]

Issue #52, opened 2026-05-04
Sandbox levels: pi


maeste:
Parent: #47 — Pattern reference: #48 (claude prototype)

> **Do not start until #48 is merged.** This task assumes the runtime dispatch logic and sandbox-policy mechanism are already in place. Reuse them.

## Scope

Apply the three-level sandbox model to **pi**, following the pattern established in #48.

## Background (self-contained)

Today: `[agents.pi]` (unsandboxed), `[agents.pi-bwrap]`, `[agents.pi-nono]`. Collapse to single `[agents.pi]` with `sandbox_level` + `sandbox_backend`.

## ⚠️ Pi-specific multi-provider complexity (CRITICAL)

Pi supports **18+ LLM providers** simultaneously. Current entries pass through env vars for: anthropic, openai, azure-openai, gemini (api+vertex), deepseek, mistral, groq, cerebras, cloudflare, xai, openrouter, ai-gateway, zai, opencode, fireworks, kimi, minimax (regular + cn), AWS Bedrock. See `agents-defaults.toml:214-246` for the full list.

This means **paranoid mode for pi cannot statically allowlist a single LLM endpoint** — the user has to declare which providers they actually use. Recommended approach:

- **paranoid**: introduce a config field (e.g., `pi_providers = ["anthropic", "openai"]`) that the plugin reads and translates to `network.credentials: [...]` for nono. Default: empty → no network at all (true paranoid). User must opt-in to specific providers.
- **normal**: today's behavior (env var passthrough)
- **permissive**: env var passthrough + github allowlist

This is the most architecturally interesting follow-up. Document the design decision clearly.

## Three levels — pi

| | paranoid | normal | permissive |
|---|---|---|---|
| nono `network` | `credentials: \$pi_providers` (user-declared subset) — empty means full block | inherited (`lince-pi.json`) | + github allowlist + credentials = full env passthrough |
| nono `filesystem` | `\$WORKDIR` rw + scratch copy of `~/.pi` | as today | + standard permissive paths |
| bwrap | `use_real_config = false` forced (verify pi equivalent — `home_ro_dirs = ["~/.pi/"]`) + no-network | as today | + standard permissive paths |
| env vars | only declared providers' keys passed (rest filtered) | full passthrough | full passthrough |

## File touch list

- `lince-dashboard/nono-profiles/lince-pi-paranoid.json` (NEW)
- `lince-dashboard/nono-profiles/lince-pi-permissive.json` (NEW)
[...]

Issue #51, opened 2026-05-04
Sandbox levels: opencode


maeste:
Parent: #47 — Pattern reference: #48 (claude prototype)

> **Do not start until #48 is merged.** This task assumes the runtime dispatch logic and sandbox-policy mechanism are already in place. Reuse them.

## Scope

Apply the three-level sandbox model to **opencode**, following the pattern established in #48.

## Background (self-contained)

Today: `[agents.opencode]` (unsandboxed), `[agents.opencode-bwrap]`, `[agents.opencode-nono]`. Collapse to single `[agents.opencode]` with `sandbox_level` + `sandbox_backend`.

## ⚠️ OpenCode-specific Bun/Landlock workaround (CRITICAL)

OpenCode is Bun-based. **Bun-based binaries crash (SIGABRT) when spawned as subprocesses under Landlock.** Current workaround in `agents-defaults.toml:185-188`:

```toml
command = ["nono", "run", "--profile", "lince-opencode", "--workdir", "{project_dir}", "--",
           "bash", "-c", "exec \"\$(dirname \"\$(readlink -f \"\$(which opencode)\")\")/.opencode\" \"\$@\"", "--"]
```

This resolves the native binary and execs it directly, bypassing the Node.js launcher's `spawnSync`. **Preserve this command shape across all three nono levels.** For bwrap, verify whether the same issue applies (likely no, since bwrap doesn't use Landlock).

## LLM provider — open question

OpenCode uses `OPENCODE_API_KEY` (custom routing) and may not be tied to a single Anthropic/OpenAI/Gemini endpoint. **Verify during execution** what API endpoints opencode actually contacts at runtime (run with debug logging, or check opencode source/docs).

If opencode talks to many providers (like pi), paranoid mode needs either:
- (a) allow only the specific provider the user has configured (require user to declare it)
- (b) allow `network.credentials: ["openai", "anthropic", "gemini"]` collectively
- (c) a permissive variant of paranoid that allows the user's set

Decide and document.

## Three levels — opencode

| | paranoid | normal | permissive |
|---|---|---|---|
| nono `network` | TBD (see provider question above) — likely allowlist openrouter/anthropic/openai endpoints | inherited (`lince-opencode.json`) | + github allowlist |
| nono `filesystem` | `\$WORKDIR` rw + scratch copy of `~/.config/opencode` | as today | + standard permissive paths |
[...]

PR #57, opened 2026-05-05
dashboard: three-level sandbox per agent (paranoid/normal/permissive) — claude prototype

### LINKED ISSUE #48: Sandbox levels: claude prototype
Parent: #47

## Scope

Implement the three-level sandbox model on **claude** as the canonical prototype. The pattern established here is then propagated to codex/gemini/opencode/pi via separate issues.

## Background (self-contained — no need to re-discuss)

LINCE today exposes `[agents.claude]` (bwrap), `[agents.claude-unsandboxed]`, `[agents.claude-nono]` as separate entries — the N-picker shows all three. The brainstorm decision: collapse to a **single** `[agents.claude]` driven by `sandbox_level` and `sandbox_backend`, with the alternative levels as commented templates so users can uncomment them without flooding the picker.

`[agents.claude-unsandboxed]` stays as a separate entry (it's an opt-out, not a level).

## Three levels — claude specifics

LLM provider for proxy/credential injection: **anthropic** (keystore account: `anthropic_api_key`).

| | paranoid | normal | permissive |
|---|---|---|---|
| nono `network` | `credentials: ["anthropic"]` (proxy LLM only, blocks the rest) | inherited from `claude-code` preset | `network_profile: "developer"` + `allow_domain: ["api.github.com","github.com","objects.githubusercontent.com"]` + `credentials: ["anthropic"]` |
| nono `filesystem` | `\$WORKDIR` rw + scratch copy of `~/.claude` (rsync pre-launch) | as today (`lince-claude.json`) | + `~/.config/gh` r, `~/.cache` r, `~/.ssh/known_hosts` r |
| bwrap `[claude]` | `use_real_config = false` forced + no-network | as today (default) | + bind-read of `~/.config/gh`, `~/.cache`, `~/.ssh/known_hosts` |
| Tools | — | — | `gh` CLI (no docker, no podman, no `git push`) |

## File touch list

**New:**
- `lince-dashboard/nono-profiles/lince-claude-paranoid.json`
- `lince-dashboard/nono-profiles/lince-claude-permissive.json`
- `sandbox/profiles/claude-paranoid.toml` (built-in, format depends on R1)
- `sandbox/profiles/claude-permissive.toml` (built-in)

**Modified:**
- `lince-dashboard/agents-defaults.toml` — single `[agents.claude]` with `sandbox_level`/`sandbox_backend` + commented templates for paranoid/permissive; remove `[agents.claude-nono]`
- `lince-dashboard/plugin/src/...` — read both attributes, dispatch to right profile/command; auto-default backend per OS
- `sandbox/agent-sandbox` — support sandbox-policy profiles (or layer on top of `[claude]` section) — see R1
- `lince-dashboard/install.sh` — copy new nono profiles, idempotent

## Acceptance criteria

- [ ] `agents-defaults.toml` has a single `[agents.claude]` with `sandbox_level`/`sandbox_backend`; `claude-nono` removed; `claude-unsandboxed` kept
- [ ] `sandbox_level = "paranoid"` on nono: `curl https://example.com` fails inside sandbox; Claude Code itself reaches Anthropic API (via proxy); modifications to `~/.claude` inside sandbox stay isolated from the real one
- [ ] `sandbox_level = "permissive"` on nono: `gh auth status` works, `gh pr create` works, `docker ps` fails, direct `git push` fails (use gh)
- [ ] N-picker shows "Claude Code" once (not three times)
- [ ] Default backend per OS works (linux=bwrap, mac=nono); explicit `sandbox_backend = "nono"` on linux works
- [ ] Manual test doc in `docs/` covering all (level × backend) cells where applicable
- [ ] `install.sh` idempotent on the new files

## Research items (resolve during execution)

- **R1** sandbox-policy profile mechanism in `agent-sandbox`: today `[profiles.*]` = LLM credential injection. Either add policy profiles or layer paranoid/permissive on top of `[claude]` section.
- **R2** Scratch copy of `~/.claude` under nono: pre-spawn hook in plugin (rsync to `\$XDG_RUNTIME_DIR/lince-<agent-id>/.claude`, set `HOME` or bind mount), or built into nono profile activation?
- **R3** bwrap `--no-network`: exists in agent-sandbox or to add?
- **R4** MCP-on-internet allowlist (Context7, Sequential, etc.) for permissive — opt-in or default?

## References

- nono networking: https://nono.sh/docs/cli/features/networking.md
- nono credentials: https://nono.sh/docs/cli/features/credential-injection.md
- Existing profile: `lince-dashboard/nono-profiles/lince-claude.json`
- Existing entry: `lince-dashboard/agents-defaults.toml:15-25`

maeste:
Closes #48. Tracked in backlog as `lince-98`. Umbrella: #47.

## What this PR does

Collapses the multiple `[agents.claude]` / `claude-bwrap` / `claude-nono` entries in `lince-dashboard/agents-defaults.toml` into a **single** entry driven by two new attributes:

- `sandbox_level = "paranoid" | "normal" | "permissive"` — default `"normal"`. **Free-form**, not a closed enum: any string resolves to a profile file by name, so users can ship custom levels without plugin changes.
- `sandbox_backend = "bwrap" | "nono"` — default per OS (linux=bwrap, mac=nono); explicit override allowed.

The N-picker now shows `Claude Code` once, not three times. Alternative levels live as commented templates in `agents-defaults.toml` so they're discoverable without cluttering the picker.

## The three levels — final threat model

| Scenario | bwrap normal | bwrap paranoid | bwrap permissive | nono paranoid |
|---|---|---|---|---|
| Leak `ANTHROPIC_API_KEY` via env dump | ❌ | ✅ proxy strips | ✅ proxy strips | ✅ keystore |
| Modify real `~/.claude` | ❌ | ✅ `use_real_config=false` | ✅ default | ✅ scratch copy |
| Open raw socket to `attacker.com` | ❌ | **✅ netns has no route** | ❌ (intentional) | ✅ Landlock |
| `curl https://attacker.com` via HTTP_PROXY | ❌ | **✅ proxy 403** | ❌ (intentional) | ✅ nono allowlist |
| `gh auth status` works | ❌ no `~/.config/gh` mounted | n/a (only Anthropic API allowed) | ✅ via mount + `GH_TOKEN` | n/a |
| `docker ps` works | depends | ❌ | ❌ (out of scope) | ❌ |

Both paranoid backends are now **kernel-enforced**:
- nono: Landlock LSM + native nono network policy
- bwrap: `--unshare-net` (fresh netns, no route) + unix-socket bridge to the credential proxy via in-sandbox `socat`. Agents see plain TCP localhost — no SDK changes needed.

## Architectural decisions worth highlighting

1. **`sandbox_level` is free-form on purpose.** Drop a `~/.config/nono/profiles/lince-claude-with-aws.json` (nono) or `~/.agent-sandbox/profiles/with-aws.toml` (bwrap), set `sandbox_level = "with-aws"`, done. No plugin recompile.

2. **`[security] allow_domains` is the per-level allowlist knob.** Append-merged, so users extend without forking fragments:

   ```toml
   # ~/.agent-sandbox/config.toml
   [security]
   allow_domains = ["pypi.org", "files.pythonhosted.org"]
   ```

   …combined with `sandbox_level = "paranoid"` gives paranoid + the two extra domains. The proxy's strict-allowlist mode auto-activates when `unshare_net = true`.

3. **bwrap paranoid path: `bwrap --unshare-net` + bind-mount unix socket + `socat` bridge.** Considered pasta initially; the deep-research-agent revealed pasta isn't a firewall by default (forwards outbound to anywhere). The unix-socket approach is simpler, kernel-enforced, and keeps TCP-localhost compatibility for all HTTP clients.

4. **MCP-on-internet for permissive is opt-in.** Users who run MCP servers reaching internet inside the sandbox extend `allow_domains` with the MCP endpoints. Out of scope for shipped defaults.

## Files

**New:**
- `lince-dashboard/nono-profiles/lince-claude-{paranoid,permissive}.json`
- `sandbox/profiles/claude-{paranoid,permissive}.toml`
- `docs/documentation/dashboard/sandbox-levels.md` (reference)
- `docs/documentation/dashboard/sandbox-levels-testing.md` (manual test plan)
- `backlog/tasks/lince-{98,99,100,101,102}*.md`

**Modified (selected):**
- `lince-dashboard/plugin/src/{config.rs,agent.rs}` — `sandbox_level` field + dispatch
- `lince-dashboard/agents-defaults.toml` — collapse claude entries
- `lince-dashboard/install.sh` — keystore setup reminder when nono is detected
- `sandbox/agent-sandbox` — `--sandbox-level` flag, fragment loader, `CredentialProxy` allow_domains + unix-socket bridge, `--unshare-net` wiring with socat wrapper, GH_TOKEN/GITHUB_TOKEN passthrough
- `sandbox/install.sh` / `sandbox/update.sh` — copy fragments, detect socat
- `lince-dashboard/README.md` — cross-link to the new docs

## Manual verification

Acceptance criteria #2–#6 require a running dashboard and are covered by the manual test plan in `docs/documentation/dashboard/sandbox-levels-testing.md`. Reviewer should run each (level × backend) cell before merging. Static AC (#1, #7–#12) verified at code-review time (parse, build, smoke tests).

## Test plan checklist

- [ ] N-picker shows `Claude Code` once
- [ ] `sandbox_level=normal` (default): existing behavior unchanged
- [ ] `sandbox_level=paranoid` on nono: arbitrary network blocked, Anthropic API works via proxy, `~/.claude` isolated
- [ ] `sandbox_level=paranoid` on bwrap: `curl https://attacker.com` fails at netns level (DNS or route error), `pgrep socat` shows one helper per running agent, Anthropic API works
- [ ] `sandbox_level=permissive` on bwrap: `gh auth status` works (file or env-token auth), `gh pr create` works, `docker ps` fails, direct `git push` fails
- [ ] Custom level via `~/.agent-sandbox/config.toml` `[security] allow_domains = [...]` extends paranoid (paranoid + pypi.org workflow)
- [ ] `install.sh` is idempotent

## Follow-ups (already filed)

- #49 #50 #51 #52 — apply the same pattern to codex / gemini / opencode / pi (depend on this PR; backlog tasks lince-99..102 carry the design context so no re-brainstorm)
- #53 — wizard 'N' picker UX changes that the single-entry-per-agent model enables
- #54 — fragment-level `extends` inheritance (so `paranoid-with-pypi.toml` can `extends = "paranoid"` instead of relying on config.toml-overlay)

### DIFF
--- a/lince-dashboard/install.sh
+++ b/lince-dashboard/install.sh
@@ -423,6 +423,19 @@ else
     [ "$HAS_NONO" = true ] && echo "  ✓ nono installed (alternative)"
 fi
 echo ""
+if [ "$HAS_NONO" = true ]; then
+    echo -e "${GREEN}Sandbox levels (paranoid/normal/permissive):${NC}"
+    echo "  Paranoid level needs the Anthropic API key in nono's keystore so"
+    echo "  the credential proxy can inject it at runtime. Populate it once:"
+    if [ "$OS_NAME" = "Darwin" ]; then
+        echo "    security add-generic-password -s nono -a anthropic_api_key -w 'sk-ant-...'"
+    else
+        echo "    secret-tool store --label 'nono anthropic' service nono account anthropic_api_key"
+    fi
+    echo "  See https://nono.sh/docs/cli/features/credential-injection.md for"
+    echo "  1Password / Apple Passwords integration."
+    echo ""
+fi
 echo -e "${GREEN}Usage:${NC}"
 echo "  source ~/.bashrc   # reload aliases"
 echo "  zd                 # launch dashboard layout"
--- a/lince-dashboard/plugin/src/agent.rs
+++ b/lince-dashboard/plugin/src/agent.rs
@@ -3,7 +3,9 @@ use std::path::PathBuf;
 
 use zellij_tile::prelude::*;
 
-use crate::config::{AgentLayout, AgentTypeConfig, DashboardConfig, DEFAULT_SANDBOX_COMMAND};
+use crate::config::{
+    shell_escape, AgentLayout, AgentTypeConfig, DashboardConfig, DEFAULT_SANDBOX_COMMAND,
+};
 use crate::sandbox_backend::SandboxBackend;
 
 /// Name of the lifecycle wrapper script for agents without native hooks.
@@ -161,19 +163,24 @@ fn synthesize_sandboxed_command(
                     // follow-ups add codex/.codex, gemini/.gemini, etc.
                     _ => return None,
                 };
+                // Sentinels use `@@…@@` rather than bare uppercase words so
+                // accidental substring overlap is impossible (e.g. a
+                // hypothetical future `AGENT_IDENTITY` placeholder would
+                // collide with `AGENT_ID` if we used bare words). With the
+                // chosen format, replacement order is irrelevant.
                 let bash = String::from(
                     "set -e; \
-                     SCRATCH=\"${XDG_RUNTIME_DIR:-/tmp}/lince-AGENT_ID\"; \
+                     SCRATCH=\"${XDG_RUNTIME_DIR:-/tmp}/lince-@@AGENT_ID@@\"; \
                      mkdir -p -- \"$SCRATCH\"; \
                      trap 'rm -rf -- \"$SCRATCH\"' EXIT; \
-                     rsync -a --delete -- \"$HOME/AGENT_HOME_SUBDIR/\" \"$SCRATCH/AGENT_HOME_SUBDIR/\"; \
-                     HOME=\"$SCRATCH\" nono run --profile NONO_PROFILE --workdir PROJECT_DIR -- INNER_COMMAND",
+                     rsync -a --delete -- \"$HOME/@@HOME_SUBDIR@@/\" \"$SCRATCH/@@HOME_SUBDIR@@/\"; \
+                     HOME=\"$SCRATCH\" nono run --profile @@NONO_PROFILE@@ --workdir @@PROJECT_DIR@@ -- @@INNER_COMMAND@@",
                 )
-                .replace("AGENT_HOME_SUBDIR", agent_home_subdir)
-                .replace("AGENT_ID", &shell_quote(agent_id))
-                .replace("NONO_PROFILE", &shell_quote(&nono_profile))
-                .replace("PROJECT_DIR", &shell_quote(project_dir))
-                .replace("INNER_COMMAND", &inner_str);
+                .replace("@@HOME_SUBDIR@@", agent_home_subdir)
+                .replace("@@AGENT_ID@@", &shell_quote(agent_id))
+                .replace("@@NONO_PROFILE@@", &shell_quote(&nono_profile))
+                .replace("@@PROJECT_DIR@@", &shell_quote(project_dir))
+                .replace("@@INNER_COMMAND@@", &inner_str);
                 vec!["bash".to_string(), "-c".to_string(), bash]
             } else {
                 let mut cmd = vec![
@@ -193,17 +200,19 @@ fn synthesize_sandboxed_command(
     })
 }
 
-/// Minimal POSIX shell single-quote escaping for embedding strings inside
-/// `bash -c '...'` wrappers. We only use this on agent-internal values
-/// (profile names, binary paths from agents-defaults.toml), never on
-/// untrusted input.
+/// POSIX shell quoting for embedding strings inside `bash -c '...'`
+/// wrappers. Strings containing only alphanumerics and a few obviously
+/// safe punctuation chars pass through unchanged for readability;
+/// anything else is wrapped in single quotes, with embedded single
+/// quotes escaped via the standard `'\''` idiom that lives in
+/// `shell_escape`.
 fn shell_quote(s: &str) -> String {
     if s.chars()
         .all(|c| c.is_ascii_alphanumeric() || matches!(c, '-' | '_' | '/' | '.' | ':'))
     {
         s.to_string()
     } else {
-        format!("'{}'", s.replace('\'', "'\\''"))
+        format!("'{}'", shell_escape(s))
     }
 }
 
@@ -275,43 +284,38 @@ fn spawn_inner(
         // (agent_type, backend, level) instead of using the static template.
         // This is what enables the single-entry-per-agent model in
         // agents-defaults.toml.
-        let cmd_template_owned;
-        let cmd_template: &[String] = if let Some(level) = &type_config.sandbox_level {
-            match synthesize_sandboxed_command(
+        // Two paths produce the agent command: synthesize_sandboxed_command
+        // bakes agent_id/project_dir directly into argv (with shell_quote
+        // for values that end up inside a bash wrapper), so its output is
+        // final and skips expand_command_template. The legacy/static path
+        // (no sandbox_level set, or agent type not yet covered) goes
+        // through the placeholder-substitution helper as before.
+        let synthesized = type_config.sandbox_level.as_deref().and_then(|level| {
+            synthesize_sandboxed_command(
                 agent_type,
                 &type_config.sandbox_backend,
                 level,
                 &id,
                 &project_dir,
-            ) {
-                Some(t) => {
-                    // Synthesized argv is already final — agent_id and
-                    // project_dir baked in (shell-quoted where they end up
-                    // inside a bash wrapper). expand_command_template still
-                    // runs over it and is a no-op since there are no
-                    // placeholders left.
-                    cmd_template_owned = t;
-                    &cmd_template_owned
-                }
-                None => {
-                    // Agent doesn't yet support sandbox_level (follow-up task).
-                    // Fall back to the static command and surface a soft warning.
-                    eprintln!(
-                        "lince-dashboard: sandbox_level={} not yet supported for agent_type={} — using static command",
-                        level, agent_type
-                    );
-                    &type_config.command
-                }
-            }
+            )
+            .or_else(|| {
+                eprintln!(
+                    "lince-dashboard: sandbox_level={} not yet supported for agent_type={} — using static command",
+                    level, agent_type
+                );
+                None
+            })
+        });
+        if let Some(argv) = synthesized {
+            expanded.extend(argv);
         } else {
-            &type_config.command
-        };
-        expanded.extend(expand_command_template(
-            cmd_template,
-            &id,
-            &project_dir,
-            profile.as_deref(),
-        ));
+            expanded.extend(expand_command_template(
+                &type_config.command,
+                &id,
+                &project_dir,
+                profile.as_deref(),
+            ));
+        }
         // For agent-sandbox (bwrap) agents, pass the profile via -P flag so
         // agent-sandbox can resolve the profile's env vars internally.
         // Nono agents already have --profile baked into their command template.
--- a/lince-dashboard/plugin/src/agent.rs
+++ b/lince-dashboard/plugin/src/agent.rs
@@ -12,7 +12,13 @@ use crate::types::{AgentInfo, AgentStatus};
 
 /// Variant suffixes stripped by `agent_type_base_name()`.
 /// Add new suffixes here when new sandbox variants are introduced.
-const AGENT_TYPE_SUFFIXES: &[&str] = &["-unsandboxed", "-bwrap", "-nono"];
+const AGENT_TYPE_SUFFIXES: &[&str] = &[
+    "-unsandboxed",
+    "-bwrap",
+    "-nono",
+    "-paranoid",
+    "-permissive",
+];
 
 /// Extract base agent type name, stripping sandbox variant suffixes.
 /// e.g. "claude-unsandboxed" → "claude", "codex-bwrap" → "codex", "gemini" → "gemini".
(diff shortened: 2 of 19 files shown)


sandbox/agent-sandbox
```diff
@@ -455,6 +485,24 @@ class _ProxyRequestHandler(BaseHTTPRequestHandler):
...
+        if self.path.startswith(("http://", "https://")):
+            parsed = urlparse(self.path)
+            dest_host = (parsed.hostname or "").lower()
+        elif host_header:
+            dest_host = host_header.split(":")[0].lower()
+
+        # Pass-through (no credential injection) for hosts in allow_domains.
+        # Used by permissive level (github.com, etc.) and by custom levels
+        # the user defines to extend paranoid with extra reachable domains.
+        if dest_host and dest_host in self.allow_domains:
```
maeste, line 521: Nit: `dest_host` is lowercased here, and `allow_domains` is lowercased at `CredentialProxy.__init__`. Consider adding a comment noting this case-insensitive contract so future contributors don't accidentally break it.

sandbox/agent-sandbox
```diff
@@ -542,6 +594,57 @@ class _ProxyRequestHandler(BaseHTTPRequestHandler):
...
+        self._bump_count()
+
+        # In HTTP_PROXY mode the agent sends an absolute URL. Otherwise we
+        # reconstruct from Host header. We don't upgrade plain HTTP to HTTPS
+        # — pass the scheme the agent asked for.
+        if self.path.startswith(("http://", "https://")):
+            upstream_url = self.path
+        else:
+            host = self.headers.get("Host", "")
+            upstream_url = f"http://{host}{self.path}"
```
maeste: Dead code path: when `self.path` is NOT an absolute URL, `dest_host` comes from the Host header (line 497). But in pass-through context, the Host header would be `127.0.0.1:PORT` (BASE_URL mode), which would never match an `allow_domains` entry like `github.com`. This branch is effectively unreachable. Consider removing it or adding a comment.

sandbox/agent-sandbox
```diff
@@ -1644,6 +1952,31 @@ def build_bwrap_cmd(config: dict, project_dir: Path,
...
+            "UNIX-CONNECT:/run/lince-proxy.sock >/dev/null 2>&1 & "
+            # Wait briefly for the socat listener to be bound before
+            # handing control to the agent (otherwise the first request
+            # races against bind() and fails with ECONNREFUSED).
+            "for _ in 1 2 3 4 5 6 7 8 9 10; do "
+            "  if (exec 3<>/dev/tcp/127.0.0.1/8118) 2>/dev/null; then "
+            "    exec 3>&-; break; "
+            "  fi; "
+            "  sleep 0.05; "
+            "done; "
```
maeste: Suggestion: add a fail-fast check after the readiness loop:

```bash
if ! (exec 3<>/dev/tcp/127.0.0.1/8118) 2>/dev/null; then
  echo "Error: socat listener failed to start" >&2; exit 1
fi
```

Without this, if socat fails to start, the agent runs with a broken proxy (ECONNREFUSED on every API call). Safe failure mode, but confusing UX.

lince-dashboard/plugin/src/agent.rs
```diff
@@ -64,6 +70,126 @@ fn expand_command_template(
...
+                // trap; SIGKILL bypasses cleanup but that's acceptable for an
+                // ephemeral scratch dir.
+                let agent_home_subdir = match base {
+                    "claude" => ".claude",
+                    // follow-ups add codex/.codex, gemini/.gemini, etc.
+                    _ => return None,
+                };
+                let bash = String::from(
+                    "set -e; \
+                     SCRATCH=\"${XDG_RUNTIME_DIR:-/tmp}/lince-{agent_id}\"; \
```
maeste: Shell injection surface (low risk): `{project_dir}` is substituted by `expand_command_template()` which does raw `.replace()`. While `NONO_PROFILE` and `INNER_COMMAND` use `shell_quote()`, `{project_dir}` at line 155 is inside double quotes but not escaped.

In practice safe: `agent_id` is always `base-N`, `project_dir` comes from local UI. But for defense-in-depth, consider shell-quoting these in the template.

lince-dashboard/plugin/src/agent.rs
```diff
@@ -64,6 +70,126 @@ fn expand_command_template(
...
+            cmd
+        }
+        SandboxBackend::Nono => {
+            let inner_str = inner_command
+                .iter()
+                .map(|s| shell_quote(s))
+                .collect::<Vec<_>>()
+                .join(" ");
+            if level == "paranoid" {
+                // Self-contained bash wrapper:
```
maeste, line 146: Nice: the nono paranoid wrapper design (scratch dir + rsync + HOME override + EXIT trap) is clean. The doc explaining why `exec` is intentionally NOT used (so bash stays alive for the trap) is a good practice.

sandbox/profiles/claude-paranoid.toml
```diff
@@ -0,0 +1,37 @@
...
+# The lists are append-merged, so the final allowlist is the base
+# (api.anthropic.com) plus your extras.
+
+[claude]
+use_real_config = false
+
+[security]
+credential_proxy = true
+unshare_net = true
+allow_domains = ["api.anthropic.com"]
```
maeste: Question: `allow_domains = ["api.anthropic.com"]` is redundant with the credential rules (which already match `api.anthropic.com`). It matters for the `enforce_allowlist` CONNECT gate and pass-through path, but could the code auto-include credential-rule domains when `enforce_allowlist` is true? That would make proxy rules the single source of truth for reachable API hosts.

sandbox/agent-sandbox
```diff
@@ -656,6 +781,93 @@ class CredentialProxy:
...
+                        pass
+
+            threading.Thread(
+                target=_forward, args=(client, upstream), daemon=True
+            ).start()
+            threading.Thread(
+                target=_forward, args=(upstream, client), daemon=True
+            ).start()
+
+        def _accept_loop() -> None:
```
maeste, line 895: The `_accept_loop` has no explicit stop condition beyond OSError on `sock.accept()`. The `stop()` method closes the socket which triggers the OSError, but consider a `threading.Event` for a cleaner shutdown pattern.

sandbox/agent-sandbox
```diff
@@ -1967,13 +2335,54 @@ def cmd_run(args, agent_extra: list[str]) -> None:
...
+            print(
+                "Error: paranoid sandbox needs 'socat' on the host (it is "
+                "bind-mounted into the sandbox to bridge TCP -> unix-socket). "
+                "Install via your package manager (e.g. `dnf install socat` "
+                "or `apt install socat`), or switch sandbox_backend to nono.",
+                file=sys.stderr,
+            )
+            sys.exit(1)
+        SANDBOX_DIR.mkdir(parents=True, exist_ok=True)
+        proxy_unix_socket = str(
```
maeste, line 2418: The unix socket path uses PID (`proxy-{os.getpid()}.sock`). If the process ever forks, the child could collide. Since `agent-sandbox` doesn't daemonize this is fine, but worth a brief comment.

lince-dashboard/install.sh
```diff
@@ -423,6 +423,19 @@ else
     [ "$HAS_NONO" = true ] && echo "  ✓ nono installed (alternative)"
 fi
 echo ""
+if [ "$HAS_NONO" = true ]; then
+    echo -e "${GREEN}Sandbox levels (paranoid/normal/permissive):${NC}"
+    echo "  Paranoid level needs the Anthropic API key in nono's keystore so"
+    echo "  the credential proxy can inject it at runtime. Populate it once:"
```
maeste, line 429: Good addition: the nono keystore setup reminder for paranoid mode. Clear instructions for both macOS and Linux.

maeste (commented):
## Hermes Agent Review

Architecturally solid PR. The free-form `sandbox_level` design is elegant and the bwrap unix-socket bridge for paranoid networking is clever. Code quality is high, security model is well-documented. A few suggestions below.

### Critical

None.

### Warnings

None.

### Suggestions

See inline comments.

### Looks Good

- The merge_sandbox_level deep-merge with append semantics for lists is clean and well-documented
- CredentialProxy unix-socket bridge design (host TCP -> unix socket -> in-sandbox socat) is a solid pattern
- shell_quote() in agent.rs for profile/command escaping
- Fragment search order (project-local -> user -> built-in) with agent-first fallback
- install.sh socat detection warning (soft fail, not hard prerequisite)
- Comprehensive test plan in sandbox-levels-testing.md

maeste:
## Code Review Summary

**Verdict: Comment** (no blocking issues, 8 suggestions)

### Critical
None.

### Warnings
None.

### Suggestions (8)
- **sandbox/agent-sandbox:502** - Add comment documenting case-insensitive comparison contract
- **sandbox/agent-sandbox:615** - Dead code path in _forward_passthrough (unreachable Host header branch)
- **sandbox/agent-sandbox:1975** - Fail-fast after socat readiness loop instead of silent broken proxy
- **agent.rs:151** - {project_dir} not shell-quoted in nono paranoid bash wrapper (low risk)
- **sandbox/profiles/claude-paranoid.toml:37** - allow_domains duplicates credential-rule domains; consider auto-inclusion
- **sandbox/agent-sandbox:847** - _accept_loop could use threading.Event for cleaner shutdown
- **sandbox/agent-sandbox:2363** - PID-based socket path could collide if process forks (comment needed)
- **lince-dashboard/install.sh:429** - Good nono keystore reminder (positive note)

### Looks Good
- Clean merge_sandbox_level deep-merge with append semantics for lists
- CredentialProxy unix-socket bridge design is solid
- shell_quote() for profile/command escaping in agent.rs
- Fragment search order (project-local, user, built-in) with agent-first fallback
- install.sh socat detection (soft fail, not hard prerequisite)
- Comprehensive test plan in sandbox-levels-testing.md
- Security scan: no hardcoded secrets, no debug leftovers, no merge conflict markers
- Port 8118 safely isolated inside --unshare-net per-instance netns
- rsync $HOME expansion correctly happens on host before nono takes over

### Security Scan
- **Secrets**: Clean - no hardcoded API keys or tokens
- **Debug statements**: Clean - no leftover print/debug/TODO
- **Merge conflicts**: Clean
- **SSRF**: _forward_passthrough gated by allow_domains + enforce_allowlist
- **Shell injection**: Low risk (local UI only, agent_id always safe, project_dir unquoted but local)

---
*Reviewed by Hermes Agent*

sandbox/agent-sandbox
maeste: Skipping this — the cost outweighs the benefit here.

The current `OSError`-on-close pattern for `_accept_loop` shutdown works correctly: `stop()` closes the listening socket, which causes the blocked `accept()` to raise, the loop catches and exits. This is the canonical way to interrupt a blocking `accept()` in stdlib without `select`/`epoll` plumbing.

Switching to `threading.Event` would require:
- An additional poll/timeout in the accept loop (so the event can be checked between accepts)
- A second cleanup path in `stop()` (set the event AND close the socket — both still needed because event alone does not unblock `accept()`)
- More state to track the bridge lifecycle

That is more code, more state, and the same actual interruption mechanism (closing the socket) underneath. The `OSError`-as-signal pattern is also what `socketserver.BaseServer.shutdown()` uses internally in the stdlib, so it is established practice.

Filing as deliberately skipped in the cleanup commit ([5f987859](https://github.com/RisorseArtificiali/lince/commit/5f987859)) along with the SandboxContext struct (one call site) and geometric backoff (~30ms cold-start win, not worth the bash complexity).

PR #117, opened 2026-05-12
feat: configurable default_agent_type for dashboard (gh#62)

### LINKED ISSUE #62: Make default agent type for 'n' (new-with-defaults) configurable
## Summary

Pressing `n` in the lince-dashboard creates a new agent using a **hardcoded** default agent type (`"claude"`, see `lince-dashboard/plugin/src/config.rs:8`). Users running multi-agent setups (codex, gemini, opencode, pi…) cannot change which agent type the `n` shortcut spawns without recompiling.

The `N` (wizard) flow already lets users pick an agent type per spawn, and `default_profile` / `default_project_dir` are already user-configurable in the dashboard config — `default_agent_type` should follow the same pattern.

## Current behavior

- `lince-dashboard/plugin/src/config.rs:8` — `pub const DEFAULT_AGENT_TYPE: &str = "claude";`
- `lince-dashboard/plugin/src/main.rs:100` — `effective_default_agent_type()` resolves to that constant (with a fallback to the first available type if `claude` isn't installed)
- `lince-dashboard/plugin/src/main.rs:473` — `n` key uses the resolved default to seed the name prompt

## Proposed change

1. Add `default_agent_type: Option<String>` to `DashboardConfig` (next to `default_profile` / `default_project_dir`) so users can set it in their dashboard TOML.
2. Update `effective_default_agent_type()` to prefer `config.default_agent_type` when set and present in `agent_types`, then fall back to the existing `DEFAULT_AGENT_TYPE` constant, then to the first available key.
3. Document the new field in the dashboard config docs / sample config.

## Acceptance criteria

- [ ] User can set `default_agent_type = "codex"` (or any registered type) in the dashboard config and `n` spawns that type.
- [ ] If the configured value isn't a registered agent type, fall back gracefully (current fallback chain) and keep the dashboard usable.
- [ ] Existing setups with no `default_agent_type` continue to default to `claude` (no behavior change).
- [ ] Docs / sample config updated.

maeste:
## Summary

- Add `default_agent_type: Option<String>` to `DashboardConfig` so users can pick which agent type the `n` shortcut spawns without recompiling.
- `effective_default_agent_type()` now consults the config first, then falls back to `"claude"`, then the existing base-match / first-available chain. Unknown values are silently ignored.
- The `N` wizard's initial agent-type selection follows the same configured base (consistent with how `default_provider` seeds the wizard's provider step).
- Docs and the shipped sample `config.toml` already advertised this field; this PR finishes the wiring on the plugin side.

## Test plan

- [x] `cargo build --target wasm32-wasip1` (debug + release) compiles cleanly.
- [x] `cargo check --tests --target wasm32-wasip1` — added two unit tests in `config.rs` covering `default_agent_type` TOML round-trip and the absent-field case.
- [ ] Manual: set `default_agent_type = "codex"` in `~/.config/lince-dashboard/config.toml`, reload the dashboard, press `n` — name prompt should default to `codex-<N>`. Press `N` — wizard should open on codex.
- [ ] Manual: set `default_agent_type = "bogus"` — `n` still works, falls back to claude. No status-bar error.
- [ ] Manual: omit the field — behavior identical to pre-PR (defaults to claude when registered).

Closes #62

### DIFF
--- a/docs/documentation/dashboard/config-reference.md
+++ b/docs/documentation/dashboard/config-reference.md
@@ -42,7 +42,7 @@ The plugin checks `config.toml` for changes every 5 seconds and applies them wit
 **Hot-reloadable** (applied immediately):
 
 - `focus_mode`, `status_method`, `max_agents`, `status_file_dir`
-- `agent_layout`, `default_provider`, `default_project_dir`
+- `agent_layout`, `default_provider`, `default_project_dir`, `default_agent_type`
 
 **Not hot-reloadable** (requires restart, only affects new agents):
 
--- a/lince-dashboard/plugin/src/config.rs
+++ b/lince-dashboard/plugin/src/config.rs
@@ -249,6 +249,12 @@ pub struct DashboardConfig {
     pub sandbox_config_path: Option<String>,
     #[serde(default)]
     pub default_project_dir: Option<String>,
+    /// Default agent type for the `n` shortcut and the `N` wizard's initial
+    /// selection. Must match a key in `agent_types` (loaded from
+    /// `agents-defaults.toml` + user overrides). Unknown values silently fall
+    /// back to `DEFAULT_AGENT_TYPE` then to the first registered type.
+    #[serde(default)]
+    pub default_agent_type: Option<String>,
     #[serde(default = "default_sandbox_command")]
     pub sandbox_command: String,
     #[serde(default)]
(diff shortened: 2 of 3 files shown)


maeste:
Closing in favor of a more substantial UX per gh#62 discussion: the static TOML `default_agent_type` field is being replaced by an in-app workflow — at the `N` wizard's final step, pressing `!` instead of Enter will spawn the agent **and** persist all wizard choices (agent_type, sandbox backend, sandbox level, provider, project_dir) as the new `n` shortcut default for this launch-dir. The default is saved alongside the open-agents state (`Q` save-and-quit). Follow-up PR coming on a new branch.

PR #131, opened 2026-05-14
fix(dashboard): make bwrap sandbox generic for any agent type

### LINKED ISSUE #132: AgentSandbox backend ignores unknown agent types, dropping --sandbox-level
## Problem

Agent types not listed in the hard-coded match in `synthesize_sandboxed_command()` (currently: `claude`, `codex`, `gemini`, `opencode`, `pi`, `bash`, `zsh`, `fish`) fall through to `_ => None`, causing the caller to use the static `command` template from `agents-defaults.toml`. That template lacks `--sandbox-level`, so the agent runs at the default `normal` level regardless of its configured `sandbox_level` (e.g. `permissive`). The permissive profile's `home_ro_dirs` (including `.ssh`) are never mounted.

This affects any custom or future agent type added to `agents-defaults.toml` without also patching the Rust match.

## Fix

Restructure `synthesize_sandboxed_command()` so the `AgentSandbox` (bwrap) backend returns early before the agent-specific match. The bwrap path only needs the base agent name to build `agent-sandbox run --agent <base> [--sandbox-level <level>]` — it doesn't use `inner_command` or `agent_home_subdir` from the match. The match (and its `_ => None` catch-all) then only applies to `Nono` and `None` backends which genuinely need per-agent knowledge.

Also expose full `~/.ssh` directory in the permissive profile instead of just `known_hosts`.

## PR

#131

paoloantinori:
## Summary

- Restructure `synthesize_sandboxed_command()` so the `AgentSandbox` (bwrap) backend returns early before the agent-specific match, making `--sandbox-level` work for **any** agent type without patching the Rust match
- Expose full `~/.ssh` directory (not just `known_hosts`) in the permissive sandbox profile

## Problem

Agent types not listed in the hard-coded match (`claude`, `codex`, `gemini`, `opencode`, `pi`, `bash`, `zsh`, `fish`) fell through to `_ => None`, which caused the caller to use the static `command` template from `agents-defaults.toml`. That template lacks `--sandbox-level`, so the agent ran at the default `normal` level regardless of its configured `sandbox_level = "permissive"`. The permissive profile's `home_ro_dirs` (including `.ssh`) were never mounted.

This affects any custom agent type added to `agents-defaults.toml` without also patching the Rust match.

## Fix

The `SandboxBackend::AgentSandbox` branch doesn't use `inner_command` or `agent_home_subdir` from the match — those are only consumed by `Nono` and `None` backends. So the agent-specific match is unnecessary for bwrap. The early return builds `agent-sandbox run --agent <base> [--sandbox-level <level>]` from just the base agent name, which agent-sandbox resolves internally.

## Test plan

- [x] Build WASM: `cd lince-dashboard/plugin && PATH="$HOME/.cargo/bin:$PATH" $HOME/.cargo/bin/cargo build --target wasm32-wasip1`
- [x] Install: `cp target/wasm32-wasip1/debug/lince-dashboard.wasm ~/.config/zellij/plugins/`
- [x] Spawn claude agent with permissive level → verify regression-free (same synthesis path, now via early return)
- [x] `ps aux | grep agent-sandbox` confirms `--sandbox-level permissive` in the bwrap command
- [x] Spawn any agent not in the hard-coded match with a sandbox level → `--sandbox-level` is now correctly passed

Fixes #132

### DIFF
--- a/lince-dashboard/plugin/src/agent.rs
+++ b/lince-dashboard/plugin/src/agent.rs
@@ -171,10 +171,33 @@ fn synthesize_sandboxed_command(
     project_dir: &str,
 ) -> Option<Vec<String>> {
     let base = agent_type_base_name(agent_type);
-    // Per-agent dispatch: (inner argv, $HOME subdir for the paranoid scratch).
+
+    // AgentSandbox (bwrap) doesn't need per-agent inner-command knowledge —
+    // agent-sandbox resolves the inner binary and args itself via --agent.
+    // Handle it generically so any agent type (including future ones) gets
+    // --sandbox-level passed through correctly.
+    if matches!(backend, SandboxBackend::AgentSandbox) {
+        let mut cmd = vec![
+            "agent-sandbox".to_string(),
+            "run".to_string(),
+            "-p".to_string(),
+            project_dir.to_string(),
+            "--id".to_string(),
+            agent_id.to_string(),
+            "--agent".to_string(),
+            base.to_string(),
+        ];
+        if level != "normal" {
+            cmd.push("--sandbox-level".to_string());
+            cmd.push(level.to_string());
+        }
+        return Some(cmd);
+    }
+
+    // Remaining backends (Nono, None) need per-agent knowledge:
+    // (inner argv, $HOME subdir for the paranoid scratch).
     // codex's `--sandbox danger-full-access` disables its own filesystem
-    // sandbox so it doesn't fight ours; the bwrap path also reads
-    // disable_inner_sandbox_args from agents-defaults.toml.
+    // sandbox so it doesn't fight ours.
     let (inner_command, agent_home_subdir): (Vec<String>, &str) = match base {
         "claude" => (
             vec![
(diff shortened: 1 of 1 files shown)


maeste:
There is conflicts

lince-dashboard/agents-defaults.toml
```diff
@@ -75,20 +75,36 @@ sandbox_level = "normal"          # paranoid | normal | permissive (or custom; s
...
-SessionStart = "input"
-UserPromptSubmit = "running"
-# Claude's `Stop` hook fires when the assistant finishes a turn and the
-# prompt returns to the user — NOT when the process exits. Map it to
-# `input`. Actual process termination is detected by the dashboard's
-# pane-exit reconciliation (agent.rs), not by a hook event.
-Stop = "input"
-idle_prompt = "input"
-permission_prompt = "permission"
+[agents.zai]
```
maeste: Does it need to be part of projects' default agents or is it your own and unintentioal included? "zai" seems your own command wrapping claude with a specifice configuration, more than anoter supported agent.

maeste (changes requested):

maeste (approved):

PR #137, opened 2026-05-14
fix(install): resolve cargo/rustc via rustup on macOS Homebrew setups

### LINKED ISSUE #133: macOS: WASM build fails with Homebrew Rust (can't find crate for `core`)
## Bug

On macOS, the installation fails during the WASM plugin build with:

```
error[E0463]: can't find crate for `core`
  = note: the `wasm32-wasip1` target may not be installed
  = help: consider downloading the target with `rustup target add wasm32-wasip1`
```

## Root cause

macOS users commonly have Rust installed via Homebrew (`brew install rust`), which provides standalone `rustc` and `cargo` at `/opt/homebrew/bin/` but **not** `rustup`. Even when rustup is installed separately via Homebrew, the standalone `cargo` may take precedence over rustup's proxy at `~/.cargo/bin/cargo`. Homebrew's cargo cannot see rustup-managed targets, so `rustup target add wasm32-wasip1` installs the target but `cargo build --target wasm32-wasip1` still fails.

This is the dual Rust installation conflict.

## Steps to reproduce

1. On macOS: `brew install rust` (and optionally `brew install rustup-init`)
2. Run `./quickstart.sh` or `./lince-dashboard/install.sh`
3. Build fails with `can't find crate for 'core'`

## Expected behavior

The install scripts should detect the Homebrew Rust conflict, offer to set up the rustup toolchain (`rustup default stable`), and fail early with clear instructions if the conflict can't be resolved automatically.

paoloantinori:
## Summary
- Fix WASM build failure on macOS when both Homebrew Rust and rustup are installed
- Resolve `cargo` and `rustc` via `rustup which` instead of relying on PATH resolution
- Set `RUSTC` env var explicitly so cargo uses the rustup-managed compiler

## Problem
On macOS with Homebrew Rust + rustup, bare `cargo` resolves to Homebrew's binary which cannot see rustup-installed targets like `wasm32-wasip1`. Even after fixing cargo resolution, cargo internally invoked Homebrew's `rustc` (no `~/.cargo/bin/rustc` proxy existed), producing the same "can't find crate for `core`" error.

Follow-up to #134, closes #133.

## Test plan
- [x] Build succeeds on Fedora (Linux)
- [x] Build succeeds on macOS with Homebrew Rust + rustup
- [ ] Full install.sh run on macOS

### DIFF
--- a/lince-dashboard/install.sh
+++ b/lince-dashboard/install.sh
@@ -134,8 +134,19 @@ if [ "$(uname -s)" = "Darwin" ]; then
     fi
     if [ -n "$BREW_CARGO" ] && ! command -v rustup >/dev/null 2>&1; then
         MISSING+=("rustup (Homebrew Rust detected at $BREW_CARGO but rustup is missing)")
-    elif [ -n "$BREW_CARGO" ] && ! rustup which cargo >/dev/null 2>&1; then
-        MISSING+=("rustup toolchain (run: rustup default stable)")
+    elif [ -n "$BREW_CARGO" ]; then
+        # rustup exists — verify the resolved cargo is actually rustup-managed,
+        # not Homebrew's standalone one.  `rustup which cargo` may succeed even
+        # when bare `cargo` still resolves to Homebrew's binary.
+        RUSTUP_CARGO="$(rustup which cargo 2>/dev/null || true)"
+        ACTIVE_CARGO="$(command -v cargo 2>/dev/null || true)"
+        if [ -z "$RUSTUP_CARGO" ]; then
+            MISSING+=("rustup toolchain (run: rustup default stable)")
+        elif [ "$RUSTUP_CARGO" != "$ACTIVE_CARGO" ] && [ ! -x "$HOME/.cargo/bin/cargo" ]; then
+            # rustup has a toolchain but the proxy in ~/.cargo/bin/ is missing —
+            # bare `cargo` will hit Homebrew's, and the build will fail.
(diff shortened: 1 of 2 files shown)


maeste:
### Code review

Found 1 issue:

1. `lince-dashboard/plugin/build.sh` removes the macOS hard-fail guard and silently falls back to bare `cargo` when `rustup which cargo` returns empty (CLAUDE.md says "Never use bare `cargo` for WASM builds … both PATH and explicit cargo path are required").

On macOS with Homebrew Rust installed but no rustup default toolchain set, `rustup which cargo` returns empty, the fallback at line 24 resolves `CARGO_BIN` to Homebrew's `cargo`, and the build proceeds with a non-rustup cargo — defeating the macOS dual-Rust protection the prior code provided with an actionable error message. The `export RUSTC` below only triggers when `RUSTC_BIN` is also set, so on this path cargo and rustc both come from Homebrew. The previous code aborted here with clear remediation steps; this fails later with the cryptic "can't find crate for \`core\`" error.

https://github.com/RisorseArtificiali/lince/blob/c6a98d8a7c388ca56b78b58728eced83b7fe6dc5/lince-dashboard/plugin/build.sh#L22-L26

Relevant CLAUDE.md guidance: https://github.com/RisorseArtificiali/lince/blob/c6a98d8a7c388ca56b78b58728eced83b7fe6dc5/CLAUDE.md#L11-L18

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #140, opened 2026-05-14
docs(multi-agent-guide): wizard step 1 reflects 2-axes design

rinaldofesta:
## Problem

`MULTI-AGENT-GUIDE.md`'s Step 1 mockup shows each agent twice (`claude` + `claude-unsandboxed`, `codex` + `codex-bwrap`, etc.). That's the pre-LINCE-98/99 design — what `base_agents()` in `config.rs` deduplicates today. A new user (or contributor) reading the guide builds a wrong mental model and is then surprised that the wizard step actually offers only the base agents.

## Fix

Rewrite the section to walk Step 1 → Step 2 → Step 3 separately:

- **Step 1 Agent Type**: base agents only; `-unsandboxed` companions are reachable through Step 2.
- **Step 2 Sandbox Backend**: `nono` / `none`; selecting `none` rewrites `effective_agent_type` to the `<base>-unsandboxed` companion and skips Step 3.
- **Step 3 Sandbox Level**: only shown for sandboxed backends; custom levels auto-discovered from `~/.config/nono/profiles/` and `~/.agent-sandbox/profiles/`.

No code changes.

PR #141, opened 2026-05-14
feat(dashboard): `U` shortcut quick-spawns an unsandboxed agent

rinaldofesta:
## Problem

The wizard's path to unsandboxed (`Shift+N` → Sandbox Backend → `none`) works, but it's hidden behind a default-tab that lands on `nono`. New users don't discover that backend `none` exists; even seasoned users hitting a greenfield bootstrap that needs `~/.npm`, `~/.cargo`, etc. have to walk the full wizard to get there.

## Fix

Add a top-level `U` keybinding that spawns the `<base>-unsandboxed` companion of the current default agent type, skipping the wizard. One-shot — does not touch `session_defaults`, so `n` keeps spawning the sandboxed configuration the user set up via `!`.

Surfaces in the help overlay and status-bar hints as `U  New-unsandboxed`. If the `<base>-unsandboxed` entry isn't configured for the current default agent, a status message explains why nothing happened.

## Why not expose `-unsandboxed` in wizard Step 1

That would revert the 2-axes design from #53 (`agent type × sandbox backend × level`, settled because the combined Cartesian product scales noisily). `U` keeps the architecture intact while offering a more discoverable path.

## Related

Surfaced while answering a user's question about why `Claude Code (unsandboxed)` doesn't appear in Step 1; same discussion produced PR #140 (doc fix for the now-obsolete mockup in `MULTI-AGENT-GUIDE.md`).

### DIFF
--- a/lince-dashboard/plugin/src/dashboard.rs
+++ b/lince-dashboard/plugin/src/dashboard.rs
@@ -699,13 +699,13 @@ fn format_key_hints_short(hints: &[KeyHint]) -> String {
 /// Return the context-appropriate key hints.
 fn status_bar_hints(empty: bool, focused: bool, detail: bool) -> Vec<KeyHint> {
     if empty {
-        vec![("n", "New-defaults"), ("N", "New-wizard"), ("Q", "Save+Quit"), ("?", "Help")]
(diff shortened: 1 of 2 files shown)


rinaldofesta:
Closing — keeping the `U` shortcut only in my local fork for now. Will reopen if maintainers want to consider it as a discoverability improvement for the unsandboxed path.

PR #144, opened 2026-05-15
fix(dashboard): poll status .state files via run_command, not std::fs

### LINKED ISSUE #143: `status_method = "file"` is dead code — `poll_status_files` reads host `/tmp` via `std::fs` from the WASI sandbox
## Labels (suggested)
`bug`, `dashboard`

## Problem

With `status_method = "file"` the dashboard never updates the Status column. `poll_status_files()` (`main.rs:1303`) runs every timer tick but always reads nothing — agents stay frozen at their boot status regardless of hook activity.

## Root Cause

`poll_status_files()` reads the per-agent `.state` files with `std::fs::read_to_string`:

```rust
let content_opt = std::fs::read_to_string(&path_claude)
    .or_else(|_| std::fs::read_to_string(&path_wrapper))
    .ok();
```

Zellij plugins run in a WASI sandbox with a **virtualized filesystem** — the host's `/tmp` is not on the plugin's view, so every `read_to_string` returns `Err` and `content_opt` is always `None`. This is the same reason the rest of the plugin reaches the host filesystem only through `run_command` (`pwd`, config-load `cat`, `detect_backend`, `complete_path`, ...). `std::fs` is unusable for host paths here.

## Fix

Read the `.state` files through a host `cat` via `run_command`, like every other host-fs access. Replace the synchronous `poll_status_files()` with an async kick-off + a result handler:

- Timer handler calls `poll_status_files_async(status_dir)` which runs
  `for f in <dir>/*.state; do printf '%s\t%s\n' "$(basename "$f" .state)" "$(tr -d '\n' < "$f")"; done`
- A new `CMD_POLL_STATUS` arm in `update()` parses the `<basename>\t<event>` lines, maps each event through the agent's `event_map`, and updates `AgentInfo.status`.

## Why it matters

`status_method` defaults to `pipe`, so most users never hit this. But `file` is the **only** working channel for a sandboxed agent whose runtime can't reach the zellij IPC socket — e.g. an agent under the `nono` backend on macOS, where the socket lives under `$TMPDIR` outside the sandbox's filesystem grants. For that setup the pipe is unreachable and the documented `file` fallback is silently broken, leaving no way to get status updates at all.

(Separately, the shipped `nono` profiles also need RW on the status dir so the hook — running *inside* the sandbox — can write the `.state` files in the first place. That's a profile-config change, tracked best as its own issue.)

## PR

(linked below)

rinaldofesta:
## Problem

With `status_method = "file"` the dashboard never updates the Status column. `poll_status_files()` runs every timer tick but reads nothing.

## Root Cause

`poll_status_files()` reads the `.state` files with `std::fs::read_to_string`. Zellij plugins run in a WASI sandbox with a virtualized filesystem — the host `/tmp` is not on the plugin's view, so every read returns `Err`. It's the same reason the rest of the plugin reaches the host filesystem only through `run_command` (`pwd`, config-load `cat`, detect_backend, complete_path).

## Fix

Read the `.state` files through a host `cat` via `run_command`:

- `config::poll_status_files_async()` runs a shell loop over `<status_dir>/*.state`, emitting `<basename>\t<event>` lines.
- A new `CMD_POLL_STATUS` arm in `update()` parses them, maps each event through the agent's `event_map`, updates `AgentInfo.status`.

## Why it matters

`status_method` defaults to `pipe`, so most users don't hit this. But `file` is the only working channel for a sandboxed agent whose runtime can't reach the zellij IPC socket — e.g. the `nono` backend on macOS, where the socket lives under `$TMPDIR` outside the sandbox's filesystem grants. There the pipe is unreachable and the documented `file` fallback is silently broken.

Closes #143.

### DIFF
--- a/lince-dashboard/plugin/src/config.rs
+++ b/lince-dashboard/plugin/src/config.rs
@@ -410,6 +410,29 @@ pub fn run_typed_command_with(args: &[&str], cmd_type: &str, extra: &[(&str, &st
 /// Command type for async path tab-completion via `run_command`.
 pub const CMD_PATH_COMPLETE: &str = "path_complete";
 
+/// Command type for async polling of per-agent `.state` status files.
+pub const CMD_POLL_STATUS: &str = "poll_status";
+
+/// Kick off async polling of the per-agent `.state` files.
+///
+/// WASI plugins have a virtualized filesystem — `std::fs` cannot see the
+/// host's `/tmp`, so the old `poll_status_files()` using `std::fs::read_to_string`
+/// silently read nothing. Read the files through a host `cat` via
+/// `run_command`, the same channel every other host-filesystem access in
+/// this plugin uses. Output is one `<basename>\t<event>` line per `.state`
+/// file (basename without the `.state` suffix); results arrive in
+/// `Event::RunCommandResult` with context `type=poll_status`.
+pub fn poll_status_files_async(status_dir: &str) {
+    let dir = shell_escape(status_dir);
+    let script = format!(
+        "for f in '{}'/*.state; do [ -f \"$f\" ] || continue; \
+         printf '%s\\t%s\\n' \"$(basename \"$f\" .state)\" \
+         \"$(tr -d '\\n' < \"$f\" 2>/dev/null)\"; done",
+        dir
+    );
+    run_typed_command(&["sh", "-c", &script], CMD_POLL_STATUS);
+}
+
 /// Command type for async discovery of custom sandbox-level profiles.
 /// Output format: lines of `<backend>:<base>:<level>` (or `<backend>:<base>:` for the
 /// suffix-less "normal" profile). See `discover_sandbox_levels_async()`.
--- a/lince-dashboard/plugin/src/main.rs
+++ b/lince-dashboard/plugin/src/main.rs
@@ -225,11 +225,12 @@ impl ZellijPlugin for State {
                     }
                 }
 
-                // File-based status polling fallback
+                // File-based status polling: kick off an async `cat` of the
+                // .state files. `std::fs` can't reach the host filesystem from
+                // the WASI plugin sandbox, so the read goes through a host
+                // shell command; results land in the CMD_POLL_STATUS handler.
                 if self.config.status_method == StatusMethod::File {
-                    if self.poll_status_files() {
-                        needs_render = true;
-                    }
+                    config::poll_status_files_async(&self.config.status_file_dir);
                 }
 
                 // Clear transient status messages after timeout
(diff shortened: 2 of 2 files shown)


maeste (commented):
## Code Review — PR #144

### Summary

This PR fixes a bug where `status_method = "file"` never updates the dashboard Status column. The root cause is that `poll_status_files()` used `std::fs::read_to_string`, which silently fails inside the WASI plugin sandbox (the host `/tmp` is not on the plugin virtualized filesystem). The fix replaces synchronous `std::fs` reads with async `run_command`-based `cat` through the host shell — the same channel every other host-filesystem access in this plugin already uses.

**Scope:** 2 files changed (+78 / -39). Clean, focused change.

---

### Quality

- **Well-scoped fix.** The PR does exactly what it says — no scope creep, no unrelated refactors.
- **Clear documentation.** Doc comments on `poll_status_files_async` and inline comments in the `CMD_POLL_STATUS` handler explain the WASI constraint and the output format. This is great for future maintainers.
- **Dual naming convention preserved.** The new handler correctly checks both `claude-{id}` and `{id}` keys (line 460–462 in the diff), matching the previous behavior.
- **Deduplication / no-change skip.** `last_polled_event` comparison is preserved, preventing unnecessary re-renders.
- **Old `poll_status_files()` cleanly removed.** No dead code left behind — just a comment pointing to the new location. Good.

---

### Security

- **`shell_escape` used for `status_dir`.** The status directory path is escaped before interpolation into the shell script. This mitigates injection via a crafted `status_file_dir` config value. ✅
- **Output parsing is defensive.** `split_once(t)` with `is_empty()` check means malformed lines are silently skipped rather than causing panics or unexpected behavior. ✅
- **`tr -d '\\n'` inside the script.** Newlines are stripped from `.state` file contents before being printed as a single TSV field. This prevents line-injection in the output format. Good defensive measure. ✅
- **Minor note:** `shell_escape` is the only guard between user-configurable `status_file_dir` and shell execution. If this function is not robust against all edge cases (e.g., embedded newlines in the path), it could be a vector. Worth verifying `shell_escape` handles `\n` in paths, but this is a pre-existing concern, not introduced by this PR.

---

### Performance

- **Single `run_command` call replaces N synchronous `std::fs` calls.** The old code called `read_to_string` twice per agent (claude + wrapper naming). The new code fires one shell `for` loop and parses all results in one callback. This is strictly better — fewer IPC round-trips and no per-agent blocking I/O.
- **HashMap for lookup.** Building a `HashMap<&str, &str>` from the output lines before iterating agents gives O(1) lookup per agent instead of O(N) scanning. Good choice at this scale.
- **No concern for large `.state` files** — these are short event strings (e.g., `"running"`, `"idle"`), not large payloads.

---

### Suggestions

1. **Empty glob handling.** The shell script uses `for f in dir/*.state; do [ -f "$f" ] || continue; ...` — this is correct and handles the `"no match"` case where the glob literal is returned. Just noting it is handled. ✅

2. **Error path for non-zero exit.** The `CMD_POLL_STATUS` arm only matches `exit_code == Some(0)`. If the `cat` command fails (e.g., `status_dir` does not exist on host), the `_ => false` fallback silently ignores the error. Consider logging a warning once so users debugging `status_method = "file"` can see that the poll is failing, rather than wondering why nothing updates (the exact same symptom as the bug this PR fixes). This is optional and not blocking.

3. **`file_events` type.** Using `HashMap<&str, &str>` with borrowed slices from `String::from_utf8_lossy(&stdout)` is correct since the borrow lives through the function. Just a note — if this function ever gets extracted or the borrow scope changes, this would need to become `HashMap<String, String>`. Not an issue now.

4. **Test coverage.** There are no tests in this PR. The logic in `poll_status_files_async` (shell script construction) and the `CMD_POLL_STATUS` parser would benefit from unit tests — especially the TSV parsing with edge cases (empty lines, lines without tabs, empty events, duplicate keys). This is understandable given the WASI/plugin constraints make integration testing hard, but unit tests on the parsing logic would be valuable as a follow-up.

---

### Verdict

**Approve.** This is a clean, well-documented fix for a real bug. The approach (async `run_command` instead of sync `std::fs`) is consistent with the rest of the codebase. No security concerns, performance is improved, and the old dead code is properly removed.

The suggestions above (logging on error, unit tests for parsing) are nice-to-haves for future follow-ups, not blockers.

---

*Reviewed by Hermes Agent*

PR #139, opened 2026-05-14
fix: macOS quickstart failures (tomlkit pip + BSD sed)

paoloantinori:
## Summary

- **tomlkit install**: `lince-config/install.sh` now tries multiple pip invocations (`python3 -m pip`, `pip`, `pip3`) before giving up — fixes Homebrew Python where `pip` isn't on PATH
- **sed compatibility**: `lince-dashboard/install.sh` alias cleanup now uses BSD-compatible `sed -i ""` on macOS instead of GNU `sed -i`

## Problems fixed

1. **tomlkit install fails** (issue #138): On macOS with Homebrew Python 3.14, `pip` is not on PATH — only `pip3` or `python3 -m pip` work
2. **sed "extra characters at end of p command"**: BSD sed (macOS) requires `sed -i ''` for in-place edits without backup; GNU sed uses `sed -i` alone

## Test plan

- [ ] Run `quickstart.sh` on macOS with Homebrew Python
- [ ] Verify tomlkit installs successfully
- [ ] Verify shell aliases are written to `.bashrc`/`.zshrc` without sed errors
- [ ] Run `quickstart.sh` on Fedora (regression check)

Closes #138

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #138, opened 2026-05-14
fix: tomlkit install fails on macOS (pip not on PATH)


paoloantinori:
## Problem

On macOS with Homebrew Python 3.14, the `lince-config/install.sh` script fails to install `tomlkit`:

```
[1/4] Checking prerequisites...
  ✓ python3 3.14.5
  tomlkit not found. Installing...
  Failed to install tomlkit. Install manually: pip install tomlkit
✗ lince-config installation failed
```

The script only tries `pip install --user tomlkit`, but on macOS:
1. `pip` may not be on PATH (Homebrew Python ships `pip3` or requires `python3 -m pip`)
2. `--break-system-packages` may be needed but isn't tried with `python3 -m pip`

## Fix

Try multiple pip invocations in order of reliability:
- `python3 -m pip` (works regardless of PATH)
- `pip` / `pip3` (fallbacks)

Each is attempted with `--user` first, then `--user --break-system-packages`.

PR #145, opened 2026-05-15
fix(dashboard): BSD sed alias setup still failing on macOS

paoloantinori:
## Summary

- Replace array-based `_sed_inplace=(sed -i "")` with explicit `if/else` branches for BSD vs GNU sed

## Problem

PR #139's array approach `(sed -i "")` still fails on macOS with the same `"extra characters at the end of p command"` error. The empty-string array element likely doesn't expand as a separate argument on macOS's bash 3.2.

## Fix

Use inline `if Darwin / else` branches with direct `sed -i ''` / `sed -i` calls — no arrays, no bash version sensitivity.

## Test plan

- [ ] Run `quickstart.sh` on macOS — alias setup step should complete without sed errors
- [ ] Run `quickstart.sh` on Fedora (regression check)

Follows up on #138.

### DIFF
--- a/lince-dashboard/install.sh
+++ b/lince-dashboard/install.sh
@@ -451,16 +451,21 @@ for rc in "$HOME/.bashrc" "$HOME/.zshrc"; do
             echo -e "${YELLOW}  Updating LINCE aliases in $(basename $rc)${NC}"
             # Remove old LINCE alias block and re-add
             # BSD sed (macOS) requires -i '' — GNU sed uses -i alone
-            _sed_inplace=(sed -i)
             if [ "$(uname -s)" = "Darwin" ]; then
-                _sed_inplace=(sed -i "")
+                sed -i '' '/# LINCE aliases/d' "$rc"
+                sed -i '' '/alias lince=/d' "$rc"
+                sed -i '' '/alias lince-floating=/d' "$rc"
+                sed -i '' '/alias zd=/d' "$rc"
+                sed -i '' '/alias z="zellij"/d' "$rc"
+                sed -i '' '/alias zn=/d' "$rc"
+            else
+                sed -i '/# LINCE aliases/d' "$rc"
+                sed -i '/alias lince=/d' "$rc"
+                sed -i '/alias lince-floating=/d' "$rc"
+                sed -i '/alias zd=/d' "$rc"
+                sed -i '/alias z="zellij"/d' "$rc"
+                sed -i '/alias zn=/d' "$rc"
             fi
-            "${_sed_inplace[@]}" '/# LINCE aliases/d' "$rc"
-            "${_sed_inplace[@]}" '/alias lince=/d' "$rc"
-            "${_sed_inplace[@]}" '/alias lince-floating=/d' "$rc"
-            "${_sed_inplace[@]}" '/alias zd=/d' "$rc"
-            "${_sed_inplace[@]}" '/alias z="zellij"/d' "$rc"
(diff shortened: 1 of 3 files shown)


lince-dashboard/install.sh
```diff
@@ -451,16 +444,21 @@ for rc in "$HOME/.bashrc" "$HOME/.zshrc"; do
             echo -e "${YELLOW}  Updating LINCE aliases in $(basename $rc)${NC}"
```
maeste, line 444: 💡 **Nitpick:** The 6 repeated sed calls could be extracted into a helper function to reduce duplication. Not blocking — the current form is explicit and easy to read.

lince-dashboard/plugin/build.sh
```diff
@@ -40,6 +37,16 @@ if [ "$(uname -s)" = "Darwin" ]; then
         echo "  Then verify: rustup which cargo" >&2
```
maeste, line 37: 💡 **Suggestion:** The macOS-only early exit when CARGO_BIN is empty is good. On Linux the PATH fallback is safe since there is no Homebrew conflict.

lince-dashboard/plugin/build.sh
```diff
@@ -54,6 +61,7 @@ fi
 
 echo "Building lince-dashboard plugin..."
 echo "  Using cargo: $CARGO_BIN"
+echo "  Using rustc: ${RUSTC:-$(command -v rustc)}"
```
maeste, line 64: ✅ Forcing RUSTC env var ensures cargo children always use the rustup-managed compiler. Root cause fix for the wasm32-wasip1 build failure on macOS + Homebrew.

maeste (commented):
## Review: PR #145 — fix(dashboard): BSD sed alias setup still failing on macOS

### Verdict: **Approve with suggestions** — clean fix, well-scoped, no blocking issues.

---

### ✅ Looks Good

- The `if Darwin / else` approach for sed is the correct solution. No array tricks, no bash version sensitivity. Straightforward and portable.
- The `RUSTC` env var + explicit `rustup which rustc` resolution in `build.sh` is a solid improvement over the fragile `cargo rustc -- --print sysroot` subprocess check.
- Error messages on macOS when cargo cannot be resolved are clear and actionable.
- The tomlkit pip install fallback chain is more robust (tries plain first, then --break-system-packages, then --user). This correctly handles the case where you are inside a virtualenv.
- Error message redirect to stderr (`>&2`) is correct practice.

### 💡 Suggestions

- **`install.sh` tomlkit loop**: The new 4-attempt pip install loop is more verbose but clearer. One minor thought: the order tries bare install first (good for venv), then --break-system-packages, then --user, then both. This is fine, but consider whether the `--user --break-system-packages` combo is ever useful. In most environments one of the earlier attempts will succeed.

maeste:
## Code Review Summary

**Verdict: Approve with suggestions** (0 blocking, 0 warnings, 3 suggestions)

### ✅ Looks Good
- `if Darwin / else` sed approach: correct, portable, no bash version tricks
- `RUSTC` env var + explicit `rustup which rustc`: solid root cause fix
- Clear error messages on macOS with actionable remediation steps
- tomlkit pip install fallback chain handles venv correctly
- stderr redirect for errors is proper practice

- 6 repeated sed calls could use a helper function (nitpick)
- tomlkit install order: `--user --break-system-packages` combo may be redundant

---
*Reviewed by Hermes Agent*

PR #156, opened 2026-05-25
fix(wizard): validate project_dir on Enter, surface error in box

### LINKED ISSUE #126: Wizard accepts malformed `project_dir`, surfaces only as cryptic backend errors
## Labels (suggested)
`bug`, `dashboard`, `wizard`, `ux`

## Problem

The Confirm step (`main.rs:776+`) passes `wizard.project_dir` straight to `agent::spawn_agent_custom()` without checking shape. Empty input, relative paths, tilde-prefixed paths, and typos all fail late inside the sandbox backend with `Failed to canonicalize path X: No such file or directory`. The error appears in the agent pane (which closes quickly) and the dashboard status bar shows only a generic spawn failure — nothing tells the user it's the field they typed.

## Fix

Validate ASAP, ideally on the `WizardStep::ProjectDir` Enter handler (`main.rs:1063+`) so the wizard simply refuses to advance to Confirm with malformed input:

```rust
let trimmed = wizard.project_dir.trim();
if trimmed.is_empty() { /* "Project dir is required." */ }
else if trimmed.starts_with('~') { /* explain the wizard doesn't expand $HOME */ }
else if !trimmed.starts_with('/') { /* "must be absolute, hint: Tab" */ }
```

Render the error inside the wizard box (a dedicated `WizardState.project_dir_error: Option<String>` works well, since `self.status_message` is hidden by the wizard overlay — see also why validating on Confirm Enter doesn't work in practice).

## Related

- #125 — completer relative-path bug; this validation also catches that case as a backstop.
- #127 — the picker's free-text fallback needs this validation too.

rinaldofesta:
Closes #126.

## Problem

The wizard's `ProjectDir` step accepts any input when Enter is pressed, even
malformed paths (relative, missing `~/...` expansion, non-existent dirs).
The failure mode is a cryptic backend error after spawn — the user has no
hint that the input was invalid until the sandbox refuses to start.

## Fix

Validate `project_dir` on Enter at the step itself, before advancing:

- Expand `~` and trim.
- Require an absolute path (`/...`). If the user typed `~/foo` and `~` did
  not expand (no `$HOME`, exotic shell), the message is explicit about
  needing the full `/Users/...` path.
- Reject paths that fail expansion with a dedicated message.
- On failure: stay on the `ProjectDir` step, render the validation error
  inline in the input box, and let the user fix with Backspace / Tab.

No state mutation on bad input — the wizard cannot leave `ProjectDir`
holding an invalid value.

## Files

- `lince-dashboard/plugin/src/main.rs` — Enter handler validation + error
  surfacing on the `ProjectDir` step.
- `lince-dashboard/plugin/src/types.rs` — `project_dir_error: Option<String>`
  on the wizard state.
- `lince-dashboard/plugin/src/dashboard.rs` — render the error line inside
  the input box.

## Test plan

- [ ] Type a relative path (`foo`) + Enter → error shown, step doesn't advance.
- [ ] Type `~/some/path` in a shell where `~` was already expanded → continues.
- [ ] Type `/non/existent/abs/path` + Enter → continues (existence is the backend's job; only shape is validated here, matching the issue scope).
- [ ] Tab still autocompletes from the same step.

### DIFF
--- a/lince-dashboard/plugin/src/dashboard.rs
+++ b/lince-dashboard/plugin/src/dashboard.rs
@@ -964,6 +964,17 @@ pub fn render_wizard(
                 push_box_line(&mut lines, "  (default: current directory)", box_width);
                 push_box_line(&mut lines, "  [Tab] autocomplete path", box_width);
             }
+            // Validation error line (set by the ProjectDir Enter handler when
+            // the path is empty/relative/tilde-prefixed). Bold red, cleared as
+            // soon as the user edits the field.
+            if let Some(ref err) = wizard.project_dir_error {
+                push_box_line(&mut lines, "", box_width);
+                push_box_line(
+                    &mut lines,
+                    &format!("  \x1b[1;31m✗ {}\x1b[0m", err),
+                    box_width,
+                );
+            }
             push_box_line(&mut lines, "", box_width);
             push_box_line(&mut lines, "  [Enter] Next  [Esc] Cancel", box_width);
         }
--- a/lince-dashboard/plugin/src/main.rs
+++ b/lince-dashboard/plugin/src/main.rs
@@ -732,6 +732,7 @@ impl State {
                     project_dir: default_dir,
                     completions: Vec::new(),
                     completion_index: None,
+                    project_dir_error: None,
                 };
                 state.step = state.active_steps().into_iter().next().unwrap_or(WizardStep::Name);
                 self.wizard = Some(state);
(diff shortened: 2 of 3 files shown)


maeste (approved):
## Summary

This PR adds client-side validation of the `ProjectDir` wizard step before advancing to `Confirm`. It introduces a new `project_dir_error` field on `WizardState`, validates on Enter (empty / tilde-unexpanded / non-absolute), renders errors inline in the wizard box, and clears them on user edits. This directly addresses #126.

**Verdict: APPROVE** — clean, well-scoped fix that matches the issue spec exactly.

---

### ✅ Correctness

- Validation logic correctly implements all three checks from #126:
  1. **Empty input** → "Project dir is required." ✓
  2. **Tilde-prefixed after failed expansion** → explicit message about using full path ✓
  3. **Non-absolute path** → "must be an absolute path" ✓
- Tilde expansion (`expand_tilde`) is attempted transparently before validation — users who type `~/foo` in environments where `$HOME` is available get seamless acceptance. When `$HOME` is unavailable (WASI sandbox), the fallback error message is clear and actionable.
- Error state is properly cleared:
  - On `Char(c)` input (line 1190) ✓
  - On `Backspace` (line 1184) ✓
  - On completion accept via Enter (line 1117) ✓
  - On successful validation advancing to next step (line 1157) ✓
- The wizard **stays** on `ProjectDir` when validation fails (`return true` at line 1155), preventing forward progress with invalid data. ✓
- Completion-accept path (when a completion is highlighted) correctly clears error and skips validation — the selected completion is already a valid absolute path from the shell completer. ✓

### ✅ Security

- No security concerns. Pure input validation against path shape — no filesystem access, no injection vectors, no secrets handling.

### ✅ Code Quality

- **Well-documented**: inline comments explain *why* validation lives here (backend requires absolute paths), *why* a dedicated field instead of `status_message` (wizard overlay hides status bar), and *why* tilde expansion is attempted. Excellent context for future maintainers.
- **Consistent style**: matches existing patterns in the file (match-on-bare-key wizard dispatch, `push_box_line` rendering).
- **Field placement**: `project_dir_error` added right after `completion_index` in `WizardState`, logically grouped with other completion/input fields.
- **Rendering**: bold red `✗` prefix with ANSI reset follows the project's existing error-display conventions.

### ⚠️ Minor Suggestions (non-blocking)

1. **No regression/unit tests for the new validation logic**. The `types.rs` module has a healthy `#[cfg(test)]` block. A few unit tests covering the validation branches (empty string, relative path, tilde-expand success/failure) would give confidence the logic stays correct as the wizard evolves. For example:
   ```rust
   // Pseudo-test idea:
   // fn test_project_dir_validation_rejects_relative_path() { ... }
   // fn test_project_dir_validation_accepts_absolute_path() { ... }
   // fn test_project_dir_validation_expands_tilde() { ... }
   ```
   These are pure functions on `WizardState` fields — no zellij host needed, same rationale as the existing tests in this module.

2. **Error message mentions `/Users/...` specifically** (line 1146). This is macOS-centric. Linux users would see `/home/...`. Consider wording like "Use the full absolute path (Tab to autocomplete)" to avoid platform assumptions. Very minor — the Tab hint immediately solves it regardless.

### ✅ Performance

- Negligible: one extra `trim()`, two `starts_with()` checks, and optional `expand_tilde()` (which is just a string prefix replacement + env var lookup). No loops, no allocations beyond the error message string.

### ✅ Testing (manual test plan from PR description)

The PR's test plan covers the key scenarios well:
- Relative path rejected ✓ (code verified)
- Tilde expansion accepted when possible ✓ (code verified)
- Non-existent absolute paths pass through ✓ (by design — existence is backend's job)
- Tab autocomplete still works ✓ (no changes to Tab handler)

---

**Overall**: Solid fix. The implementation is clean, well-commented, and precisely scoped to the issue. The only meaningful gap is the absence of unit tests for the new validation branches, which I'd suggest adding in a follow-up.

Issue #127, opened 2026-05-13
[RFC] Smart Project Picker for the wizard's Project dir step


rinaldofesta:
## Labels (suggested)
`enhancement`, `dashboard`, `ux`, `rfc`

## Problem

The wizard's Project dir step (`main.rs:1063+`) is the only step that uses a free-text input — every other step is a navigable list (`j/k` + Enter). Three pain points compound:

1. The field is empty on first use, so the user has to type a full absolute path (35-50 chars on macOS).
2. Zellij plugin WASM hosts deliver keystrokes as `BareKey::Char(c)` one at a time, so `Cmd+V` paste is dropped. The user has to type each character.
3. Tab completion has a relative-path bug (#125).

## Proposed UX

Replace the text input with a navigable list seeded from recently-used `project_dir` values (extracted from the persisted state), with a free-text fallback for the long tail. Consistent with the other wizard steps.

```
┌─ Step 6/7: Project dir ───────────────────────────────┐
│ Filter:                                               │
│                                                       │
│ RECENTS                                               │
│ > /Users/me/Projects/Personal/synoptic    (just now)  │
│   /Users/me/Projects/Personal/hermes      (2h ago)    │
│   /Users/me/Work/cliente-x-api            (yesterday) │
│                                                       │
│ [type]: filter   [Enter]: select   [i]: free text     │
└───────────────────────────────────────────────────────┘
```

Typing filters in place (substring + subsequence, case-insensitive). `i` switches to free-text input (current behavior preserved as escape hatch).

## Design

**State.** Extend `WizardState` (`types.rs:131`) with `available_project_dirs: Vec<RecentProject>`, `project_dir_index: usize`, `project_dir_filter: String`, `project_dir_mode: { List, Input }`. Mirrors the existing AgentType/SandboxBackend/SandboxLevel/Provider pattern.

**Storage.** Extend `SavedState` (`state_file.rs`, schema v3→v4) with `global_recent_project_dirs: Vec<RecentProject>` and a `record_project_use(path)` helper called from `spawn_agent_custom`. Backward-compatible via `#[serde(default)]`, same pattern v1→v2→v3 used.

**Filter.** In-house ~40 LOC: case-insensitive substring + subsequence with a small score boost for path-segment-start matches. No new crate — consistent with the current minimal Cargo.toml (`zellij-tile`, `serde`, `serde_json`, `toml`).

[...]

Issue #125, opened 2026-05-13
Wizard completer returns relative paths, breaking sandbox spawn


rinaldofesta:
## Labels (suggested)
`bug`, `dashboard`, `wizard`

## Problem

`complete_path_async()` in `config.rs:720` runs the directory glob via `sh -c` in the host's working directory and writes the matched basename into `wizard.project_dir` verbatim. When the user types a bare prefix like `syn` and hits Tab, the field gets filled with `synoptic/` — a path the spawn step then passes to nono / agent-sandbox as `--workdir synoptic/`, which fails with `Failed to canonicalize path synoptic/: No such file or directory`.

Absolute and tilde-prefixed input work because the matched output is already absolute. Only the bare-name case is broken.

## Root Cause

The handler at `main.rs:411-440` assigns the match to `wizard.project_dir` unchanged. `complete_path_async()` itself emits whatever `for d in <prefix>*/` produces, which is relative when the prefix is relative.

## Fix

Normalize the match against `self.launch_dir` before writing it to `wizard.project_dir`. Two-line closure in the `CMD_PATH_COMPLETE` handler:

```rust
let absolutize = |p: String| -> String {
    if p.starts_with('/') { return p; }
    match &self.launch_dir {
        Some(base) => format!("{}/{}", base.trim_end_matches('/'), p),
        None => p,
    }
};
```

Apply to both the single-match and `common_prefix` branches.

## Related

- #126 — wizard should also validate the field before spawn, as a backstop against any future regression here.
- #127 — the Smart Project Picker RFC assumes this bug is fixed for the free-text fallback.

maeste:
Addressed by #208 (merged): `lince-config discover` emits absolute, existence-validated paths only (asserted by scripts/tests/test_discover.py — "all emitted paths are absolute"), which is the discovery surface the wizard/skill consume. The relative-path class is fixed at the source. Closing; reopen if the wizard's own completer still surfaces a relative path independent of discover.

rinaldofesta:
@maeste reopen this one (GitHub won't let me — you closed it): #208 fixed the `lince-config discover` surface, but the bug this issue reports is in the wizard's own Tab completer, and it's still on main @ `9251301`:

- `project_search_roots` defaults to empty (`plugin/src/config.rs:367`), so bare-name input takes the legacy glob branch (`config.rs:887-894`), which emits **relative** `dir/` names from the launch CWD.
[...]

PR #157, opened 2026-05-25
fix(wizard): normalize trailing-slash on completer matches + swimlane key

rinaldofesta:
Related to #125.

## Problem

The wizard's path completer historically returns directories with a trailing
`/` (the legacy `for d in <prefix>*/; do echo "\$d"; done` shell glob). When
the multi-root recursive `find` was introduced (RFC #127), it returns paths
without trailing slash. Two agents pointed at the same project but spawned
through different code paths end up with `project_dir` values like
`/Users/me/proj/synoptic/` vs `/Users/me/proj/synoptic` — and the dashboard
groups agents by `project_dir` string, so they fall into distinct swimlanes.

Issue #125 also tracks the broader relative-path bug; this PR addresses
the trailing-slash discrepancy that surfaces alongside it (sandbox spawn
also gets unhappy with a path that ends in `/`).

## Fix

- Normalize completer output: strip trailing `/` before storing in
  `wizard.project_dir`.
- Normalize the swimlane group key in `sort_agents_by_dir` (function
  `workdir_key`) so an agent with `…/proj` and one with `…/proj/` land
  in the same lane regardless of where the value originated.

The swimlane normalization is a comparison-only helper — it does NOT
mutate the stored `agent.project_dir`, so the value the user typed is
preserved for display.

## Files

- `lince-dashboard/plugin/src/main.rs` — strip trailing slash on completer
  result; `workdir_key` helper for swimlane grouping.
- `lince-dashboard/plugin/src/dashboard.rs` — use `workdir_key` when
  computing the swimlane group.

## Test plan

- [ ] Spawn agent A with `~/Projects/foo` (no slash). Spawn agent B with
      `~/Projects/foo/` (trailing slash). They appear in the **same** swimlane.
- [ ] Use Tab autocomplete on a partial → completed value has no trailing slash.
- [ ] Backend spawn does not receive `…/foo/` paths.

### DIFF
--- a/lince-dashboard/plugin/src/dashboard.rs
+++ b/lince-dashboard/plugin/src/dashboard.rs
@@ -376,10 +376,12 @@ fn render_agent_table(
     let mut header_dirs: Vec<String> = Vec::new();
     let mut last_dir: Option<&str> = None;
     for (i, agent) in agents.iter().enumerate() {
-        if last_dir != Some(&agent.project_dir) {
+        // Compare/display normalized so legacy `path/` and new `path` group together.
+        let key = agent.project_dir.trim_end_matches('/');
+        if last_dir != Some(key) {
             virtual_rows.push(None);
-            header_dirs.push(agent.project_dir.clone());
-            last_dir = Some(&agent.project_dir);
+            header_dirs.push(key.to_string());
+            last_dir = Some(key);
         }
         virtual_rows.push(Some(i));
         header_dirs.push(String::new());
--- a/lince-dashboard/plugin/src/main.rs
+++ b/lince-dashboard/plugin/src/main.rs
@@ -420,9 +420,18 @@ impl ZellijPlugin for State {
                             let mut matches: Vec<String> = output
                                 .lines()
                                 .filter(|l| !l.is_empty())
-                                .map(|l| config::collapse_tilde(l))
+                                // Normalize the trailing-slash inconsistency
+                                // between the legacy glob path completer (which
+                                // returns `path/`) and the multi-root `find`
+                                // (which returns `path`). Otherwise two agents
+                                // pointed at the same project but spawned from
+                                // different code paths end up in distinct
+                                // swimlanes — see also `sort_agents_by_dir`.
+                                .map(|l| l.trim_end_matches('/').to_string())
+                                .map(|l| config::collapse_tilde(&l))
                                 .collect();
                             matches.sort();
+                            matches.dedup();
 
                             if matches.len() == 1 {
                                 // Single match: auto-fill directly.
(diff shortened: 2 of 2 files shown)


maeste (approved):
## Review: PR #157 – normalize trailing-slash on completer matches + swimlane key

### Verdict: **Approve** (with minor suggestions)

---

### What this PR does

Three-layer defense against trailing-slash inconsistency in `project_dir` values:

1. **Source normalization** (`main.rs:430`): Strip trailing `/` from path-completer results *before* storing into `wizard.project_dir`. This prevents the slash from entering agent state in the first place.
2. **Sort normalization** (`main.rs:611-612,622`): New `workdir_key()` helper used in `sort_agents_by_dir()`, so agents whose `project_dir` differs only by a trailing slash sort into the same group.
3. **Display normalization** (`dashboard.rs:380`): Swimlane header grouping also trims trailing slashes, so legacy agents (stored with `/`) and new agents (without) render under the same header.

Bonus: `matches.dedup()` after sort removes duplicate entries that can appear when both `path/` and `path` were present in the raw completer output.

---

### Checklist

- **Correctness**: PASS — Logic is sound. `trim_end_matches('/')` is the standard Rust idiom for this.
- **collapse_tilde signature change**: PASS — `.to_string()` yields String; passing `&l` gives `&str` which matches `fn collapse_tilde(path: &str)`. No issue.
- **dedup() placement**: PASS — Called after `sort()`; correct since `dedup()` requires sorted input.
- **workdir_key purity**: PASS — Pure function, no side effects, doesn't mutate agent.project_dir. Good doc comment explaining intent.
- **Existing pattern consistency**: PASS — Same `trim_end_matches('/')` already used in state_file.rs:13 and config.rs:584. Consistent with codebase conventions.
- **Scope creep**: PASS — Only touches the two files described. No unrelated changes.
- **PR description quality**: PASS — Clear problem statement, explains both root causes (legacy glob vs multi-root find), links to #125, has test plan.
- **CI / checks**: WARNING — No CI status reported yet (empty rollup). Recommend confirming build passes before merge.

---

### Minor suggestions (non-blocking)

**1. Duplicated normalization logic across modules**

dashboard.rs:380 does inline `agent.project_dir.trim_end_matches('/')` while main.rs:622 calls `workdir_key()`. Both do the same thing, but workdir_key is fn-level private in main.rs so dashboard.rs can't reach it. This is fine today, but if the normalization rule ever grows (e.g., handling double-slashes, case-insensitive filesystems), you would need to update both sites. Consider extracting to a shared utils or config module if more callers appear.

**2. Edge case: root path /**

`"/".trim_end_matches('/')` yields `""`. If somehow an agent's project_dir is `/`, the swimlane key becomes an empty string and the header would show nothing (or `~` after collapse_tilde). This is vanishingly unlikely in practice (nobody runs agents from filesystem root), but worth noting.

**3. Test plan checkboxes are unchecked**

The PR body lists manual test steps but none are checked. Suggest verifying at least the 'same swimlane' scenario before merging, since it is the core regression being fixed.

PR #158, opened 2026-05-25
feat(wizard): zoxide-like recursive Tab completion across project_search_roots

rinaldofesta:
Implements #127 (RFC: Smart Project Picker).

## Problem

The wizard's Project dir Tab completion only globs the launch directory.
A user with projects scattered across `~/Projects/Personal/<name>`,
`~/code/<client>/<repo>` etc. has to type the full path every time.

## Approach

Add two opt-in config keys under `[dashboard]`:

```toml
project_search_roots = [
    "/Users/me/Projects",
    "/Users/me/code",
]
project_search_max_depth = 3   # optional, default 3
```

When the user types a **bare name** (no `/`, no `~/`) and at least one
search root is configured, Tab triggers a depth-limited `find` across
every root and returns matching directories by basename.

Example: typing `syn` + Tab in the wizard finds
`/Users/me/Projects/Personal/synoptic` regardless of where lince was
launched from. Single match → auto-fill; multiple → fill the common
prefix and show candidates.

## Backward compatibility

The default is `project_search_roots = []`, which preserves the legacy
glob-in-launch-dir behavior exactly. Existing users are unaffected
until they opt in. Absolute / tilde prefixes always use the legacy
glob branch (power users who type a specific path keep the familiar
behavior).

## Performance

`find` is depth-limited (default 3, configurable) and prunes the usual
noisy subtrees (`node_modules`, `.git`, `.cache`, `target`). Latency
stays under ~100ms on a populated `$HOME` in my testing.

## Files

- `lince-dashboard/plugin/src/config.rs` — `project_search_roots`,
  `project_search_max_depth` fields; updated `complete_path_async` to
  dispatch by input shape (absolute/tilde → legacy glob; bare name +
  roots → multi-root `find`; bare name + no roots → legacy CWD glob).
- `lince-dashboard/plugin/src/main.rs` — `BareKey::Tab` handler passes
  the two new config values into `complete_path_async`.

## Test plan

- [ ] Configure `project_search_roots = ["/path/to/projects"]` in config.toml.
- [ ] Open wizard → ProjectDir step. Type a partial that exists deep in the root → Tab autocompletes.
- [ ] Configure empty list (legacy) → Tab still works in launch dir as before.
- [ ] Type `/abs/path` + Tab → uses legacy glob unchanged.
- [ ] Type `~/foo` + Tab → uses legacy glob unchanged.

### DIFF
--- a/lince-dashboard/plugin/src/config.rs
+++ b/lince-dashboard/plugin/src/config.rs
@@ -207,6 +207,11 @@ impl SandboxColors {
     }
 }
 
+fn default_project_search_max_depth() -> usize {
+    3
+}
+
+
 fn default_status_file_dir() -> String {
     "/tmp/lince-dashboard".to_string()
 }
@@ -283,6 +288,17 @@ pub struct DashboardConfig {
     /// in `~/.config/lince-dashboard/config.toml`.
     #[serde(default)]
     pub sandbox_colors: SandboxColors,
+    /// Root directories scanned recursively when the wizard's Project dir
+    /// Tab-completion receives a bare name (no `/` or `~/` prefix). Enables
+    /// zoxide-like UX: `syn` + Tab finds `<root>/Personal/synoptic` anywhere
+    /// inside any configured root. Empty list = legacy glob in launch_dir.
+    #[serde(default)]
+    pub project_search_roots: Vec<String>,
+    /// Maximum depth (passed to `find -maxdepth`) when scanning
+    /// `project_search_roots`. Default 3 covers typical category-level
+    /// nesting (e.g. `Projects/Personal/<proj>`, `Projects/Work/<client>/<repo>`).
+    #[serde(default = "default_project_search_max_depth")]
+    pub project_search_max_depth: usize,
     /// Custom sandbox levels discovered from the filesystem, keyed by
     /// `<backend>:<base>` (e.g. `"nono:claude"` or `"agent-sandbox:claude"`).
     /// Populated asynchronously by `discover_sandbox_levels_async()` reading
@@ -311,6 +327,8 @@ impl Default for DashboardConfig {
             agent_types: HashMap::new(),
             sandbox_backend: BackendConfig::default(),
             sandbox_colors: SandboxColors::default(),
+            project_search_roots: Vec::new(),
+            project_search_max_depth: default_project_search_max_depth(),
             discovered_sandbox_levels: HashMap::new(),
         }
     }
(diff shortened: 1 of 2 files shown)


maeste (commented):
## Summary

This PR implements a **zoxide-like recursive Tab completion** for the wizard's Project dir field, addressing the core pain point from RFC #127 (typing full paths in a WASM text input). Two new opt-in `[dashboard]` config keys drive a multi-root `find` scan when the user types a bare name and presses Tab.

**Verdict: ✅ Approve with suggestions** — Clean, well-structured, backward-compatible.

---

### ✅ What's done well

1. **Backward compatibility is correct.** `project_search_roots` defaults to `Vec::new()`, so existing users hit the legacy CWD-glob branch unchanged. Absolute/`~/` prefixes always take the legacy path too. Zero-surprise opt-in.

2. **Dispatch logic is clean.** Four cases (empty → no-op, absolute/tilde → legacy glob, bare+roots → multi-root find, bare+no-roots → legacy glob) are clearly separated and documented inline.

3. **Stale-result guard preserved.** The `prefix` context key already existed and continues to work — if the user keeps typing while a scan is in flight, the stale result is discarded.

4. **Defensive shell scripting.**
   - User input (`trimmed`) goes through `shell_escape()` before interpolation into single-quoted `-name '{pattern}'` — **injection-safe**.
   - Root paths go through `shell_path_expr()` (consistent with every other host-FS access in the plugin).
   - Each `find` has its own `2>/dev/null` so a missing root doesn't spam stderr.
   - `max_depth.max(1)` prevents `find -maxdepth 0` footgun.
   - `sort -u | head -n 50` deduplicates and caps output.

5. **Noise pruning.** `node_modules`, `.git`, `.cache`, `target` exclusions are sensible defaults for dev homedirs.

6. **Side benefit: sidesteps #125.** Because the multi-root `find` returns **absolute paths**, the relative-path bug in the completer is avoided entirely when using search roots. Nice.

7. **Serde defaults are consistent with the rest of the file.** Custom `default_project_search_max_depth()` fn follows the same pattern as `default_status_file_dir()`, etc.

---

**S1. DRY: extract the legacy glob script into a helper**

The glob branch is duplicated three times (absolute/tilde, bare+no-roots, and structurally similar to the old code). A one-liner helper would reduce copy-paste drift risk:

```rust
fn glob_dirs_script(prefix: &str) -> String {
    format!(
        "for d in {}*/; do [ -d \"$d\" ] && echo \"$d\"; done 2>/dev/null | head -n {}",
        shell_path_expr(prefix), MAX_COMPLETIONS
    )
}
```

Not a blocker — the three copies are all on 4 lines and unlikely to diverge — but worth a cleanup if you touch this area again.

**S2. Consider logging when `project_search_roots` contains non-existent paths**

If a root doesn't exist or isn't a directory, every Tab press silently fires a `find` that returns nothing. Users may wonder why completion "doesn't work." A one-time `status_message` warning on first use (detectable because `completions.is_empty()` after the result arrives) would improve debuggability:

```
⚠ No projects found under configured search roots. Check [dashboard].project_search_roots.
```

Low priority — the feature is opt-in and power users who configure this can figure it out.

**S3. Test plan checkboxes are all unchecked**

The PR description has a well-thought-out test plan with 5 items, all `[ ]`. Even manual testing notes (e.g., "tested with depth=3 on a dir with 200+ subdirs, got 45ms") would give reviewers confidence this was exercised beyond "it compiles."

---
### ⚠️ Minor observations (no action needed)

1. **RFC #127 divergence (informational)**: RFC #127 proposed a *recents-based navigable list* with free-text fallback. This PR takes a different (and arguably simpler) approach: recursive `find` across configured roots. Both solve the same problem; this one trades persistence/recency for zero storage changes. Valid trade-off, worth documenting why this direction was chosen over the RFC's original proposal (a sentence in the commit body suffices).

2. **`find | head` SIGPIPE behavior**: When `head -n 50` closes the pipe after reading 50 lines, `find` receives `SIGPIPE` and exits. This is correct and efficient — the scan stops early. Just confirming this is intentional (it is).

3. **`-not -path '*/node_modules/*'` semantics**: This excludes any match whose **full path** contains `/node_modules/` as a component. It won't accidentally exclude a project literally named `my-node-modules-project` (because the pattern requires directory separators around the name). Correct behavior confirmed.

---
### 🔒 Security

**Shell injection analysis: PASSED**
- User-typed `partial` → `shell_escape()` → single-quoted inside `-name '...'` ✅
- Config-sourced `root` → `shell_path_expr()` → `'$HOME...'` or `'<escaped>'` ✅
- `max_depth` is a `usize` (integer, controlled via TOML) interpolated via `{}` format arg — not user-typed at runtime ✅

No vulnerabilities introduced.

Issue #108, opened 2026-05-11
Config-preserving merge across module update.sh scripts


maeste:
## TLDR

Stop `lince update` from destroying user-customized TOML configs. Introduces a shared `tomlkit`-based merge utility (`lince-cli/lib/config_merge.py`) and rewires each module's `update.sh` to preserve user values while still adding any new keys shipped in the new defaults. Adds a `--reset-config` opt-out on the `lince update` CLI.

## Why

Today's update scripts are destructive:

- `lince-dashboard/update.sh:55-60` — backs up `~/.config/lince-dashboard/config.toml` then **overwrites** it from defaults. User edits are lost; recovery requires manually diffing `.bak.*` files.
- `sandbox/update.sh:37` — overwrites `~/.agent-sandbox/agents-defaults.toml` with no backup.
- `sandbox/update.sh:46-58` — overwrites `~/.agent-sandbox/profiles/*.toml` likewise.

The first `lince update` shipped via npm would silently wipe out customizations users have accumulated under the old curl-only flow. That's a regression we must prevent before #102 (`lince update`) goes live.

Good news: precedent exists. `lince-config` already uses `tomlkit` for surgical format-preserving edits (line 35 import, `_dotted_set/_dotted_unset` at lines 254-282). `sandbox/update.sh:94-120` already does a narrow regex-based merge for env-passthrough. The new utility generalises these patterns.

## Implementation plan

1. **Create the merge utility** — `/home/maeste/project/lince/lince-cli/lib/config_merge.py`:
   - Uses `tomlkit` (already a project dep).
   - API: `merge_config(user_path: Path, new_default_path: Path, *, dry_run: bool=False) -> MergeResult` where `MergeResult.orphans: list[str]` lists user keys absent from new defaults.
   - Semantics:
     - **User wins** for any key present in both files.
     - **New defaults added** only where missing in user file.
     - **Orphan user keys preserved** AND emitted via `MergeResult.orphans` so the calling `update.sh` can report them.
     - Comments and key ordering preserved (tomlkit's job).
     - Always writes a `<path>.bak.<YYYYmmdd-HHMMSS>` backup before mutating the user file.
     - Fail-safe on malformed input — don't write, exit non-zero with clear stderr.
   - Also callable as a script: `python3 -m lince_cli.config_merge <user> <default>` so bash `update.sh` can invoke it without Python boilerplate.

[...]

Issue #102, opened 2026-05-11
Implement `lince update` (npm-based)


maeste:
## TLDR

Implement the real `lince update` subcommand. It checks `npm` is available, runs `npm install -g @risorseartificiali/lince-cli@latest`, detects which modules are installed, and runs each module's existing `update.sh` from the freshly installed npm package tree.

## Why

This is the headline feature of the epic: replace the "re-run the curl one-liner to upgrade" UX with a single `lince update` command that's discoverable, predictable, and safe. By delegating to each module's existing `update.sh`, we avoid duplicating any install logic. By going through npm, we get version pinning, prebuilt artifacts (no `cargo`/`rustup` on user machines), and a real distribution channel.

## Implementation plan

1. **Replace the `cmd_update` stub** in `lince-cli/lince` with the real implementation.
2. **Pre-flight check**: `shutil.which("npm")` — if missing, print:
   ```
   npm not found. Lince now distributes via npm. Install Node.js from https://nodejs.org/
   then run `lince update` again.
   ```
   and exit non-zero.
3. **`--check` mode**: run `npm view @risorseartificiali/lince-cli version` (capture stdout). Compare to the version embedded in the currently-installed `lince` binary (add a `__version__` constant if not present). Print "up to date" or "update available: X → Y". Do not mutate. Exit 0.
4. **Update flow**:
   - Run `npm install -g @risorseartificiali/lince-cli@latest` with stdout streamed; raise on non-zero exit.
   - Resolve `npm root -g` → `<root>/@risorseartificiali/lince-cli/dist/`.
   - **Module detection** (filesystem-based, no manifest):
     - `sandbox`: `~/.local/share/lince/bin/agent-sandbox` exists.
     - `voxcode`: `shutil.which("voxcode") is not None`.
     - `lince-dashboard`: `~/.config/lince-dashboard/config.toml` exists.
     - `lince-cli` / `lince-config`: always (the running binary is the proof).
   - If `--module <name>` was specified (repeatable), restrict to that set; otherwise update all detected.
   - **Run updates in order**: `sandbox` → `voxcode` → `lince-dashboard` → `lince-config` → `lince-cli`. For each, `subprocess.run(["bash", str(dist_path/module/"update.sh")], check=True)` with streamed output. **Stop on first failure** with module name and the stderr tail.
[...]

PR #159, opened 2026-05-25
fix(update): preserve user config.toml and agents-defaults.toml

rinaldofesta:
## Problem

`lince-dashboard/update.sh` destroyed user customizations on every run:

- **`config.toml`**: backed up with a timestamp, then **always overwritten** from the shipped template — silently discarding API keys, security levels, agent tweaks, etc.
- **`agents-defaults.toml`**: **always copied** over the user's file, with no backup at all.

A user who ran `update.sh` after any release lost their local configuration.

## Fix

Preserve existing user files, and ship the upstream defaults as a `.dist` sidecar so new keys stay discoverable without clobbering customizations:

- `config.toml` / `agents-defaults.toml`: written **only** when the destination does not already exist (first install behaves exactly as before).
- On every run, the shipped defaults are (re)written to a sibling **`*.dist`** file:
  - `~/.config/lince-dashboard/config.toml.dist`
  - `~/.config/lince-dashboard/agents-defaults.toml.dist`
- Users can review new upstream keys with:
  ```bash
  diff -u ~/.config/lince-dashboard/config.toml.dist ~/.config/lince-dashboard/config.toml
  ```

Non-config assets (plugin WASM, layouts, hooks, wrappers, skills) keep being replaced on every update — correct distinction between user-owned and repo-owned files.

## Why the `.dist` sidecar (addressing review feedback)

@maeste rightly flagged that a pure "skip if exists" guard introduces the opposite problem: a user with an existing `config.toml` would **never** receive new default keys from future releases, and "delete it manually" is a destructive workaround.

The `.dist` sidecar is the low-effort option from that discussion: it surfaces every new upstream key via a plain `diff`, with zero risk to user data and no extra tooling. Automatic key-merging (tomlkit-based) is intentionally **out of scope** here and tracked as follow-up in #108.

## Test plan

- [x] Existing `config.toml` is preserved (not overwritten, not backed up away)
- [x] Existing `agents-defaults.toml` is preserved
- [x] `config.toml.dist` and `agents-defaults.toml.dist` are written with current upstream defaults on every run
- [x] First-time install (no existing files) still receives the shipped defaults
- [x] `bash -n update.sh` passes; verified end-to-end by running `update.sh` against a populated config dir

### DIFF
--- a/lince-dashboard/update.sh
+++ b/lince-dashboard/update.sh
@@ -52,13 +52,12 @@ CONFIG_DIR="$HOME/.config/lince-dashboard"
 CONFIG_DST="$CONFIG_DIR/config.toml"
 mkdir -p "$CONFIG_DIR"
 
-if [ -f "$CONFIG_DST" ]; then
-    BACKUP="${CONFIG_DST}.bak.$(date +%Y%m%d_%H%M%S)"
-    cp "$CONFIG_DST" "$BACKUP"
-    echo -e "${YELLOW}  Existing config backed up → $(basename "$BACKUP")${NC}"
+if [ ! -f "$CONFIG_DST" ]; then
+    cp "$SCRIPT_DIR/config.toml" "$CONFIG_DST"
+    echo -e "${GREEN}  ✓ Config installed${NC}"
+else
+    echo -e "${GREEN}  ✓ Existing config preserved (delete it manually to receive new defaults)${NC}"
 fi
-cp "$SCRIPT_DIR/config.toml" "$CONFIG_DST"
-echo -e "${GREEN}  ✓ Config updated${NC}"
 echo ""
 
 # ── Hooks ──────────────────────────────────────────────────────────────
@@ -80,9 +79,11 @@ echo -e "${GREEN}[7/8] Updating agent defaults...${NC}"
 AGENTS_DEFAULTS_SRC="$SCRIPT_DIR/agents-defaults.toml"
 AGENTS_DEFAULTS_DST="$CONFIG_DIR/agents-defaults.toml"
 
-if [ -f "$AGENTS_DEFAULTS_SRC" ]; then
+if [ -f "$AGENTS_DEFAULTS_SRC" ] && [ ! -f "$AGENTS_DEFAULTS_DST" ]; then
     cp "$AGENTS_DEFAULTS_SRC" "$AGENTS_DEFAULTS_DST"
-    echo -e "${GREEN}  ✓ Agent defaults updated${NC}"
+    echo -e "${GREEN}  ✓ Agent defaults installed${NC}"
+elif [ -f "$AGENTS_DEFAULTS_DST" ]; then
+    echo -e "${GREEN}  ✓ Existing agent defaults preserved (delete it manually to receive new upstream defaults)${NC}"
 fi
 echo ""
 
--- a/lince-dashboard/update.sh
+++ b/lince-dashboard/update.sh
@@ -56,8 +56,13 @@ if [ ! -f "$CONFIG_DST" ]; then
     cp "$SCRIPT_DIR/config.toml" "$CONFIG_DST"
     echo -e "${GREEN}  ✓ Config installed${NC}"
 else
-    echo -e "${GREEN}  ✓ Existing config preserved (delete it manually to receive new defaults)${NC}"
+    echo -e "${GREEN}  ✓ Existing config preserved${NC}"
 fi
+# Always refresh the shipped-defaults sidecar so users can diff against new
+# upstream keys without losing their customizations:
+#   diff -u "$CONFIG_DST.dist" "$CONFIG_DST"
+cp "$SCRIPT_DIR/config.toml" "$CONFIG_DST.dist"
+echo -e "${GREEN}  ✓ Upstream defaults available at config.toml.dist${NC}"
 echo ""
 
 # ── Hooks ──────────────────────────────────────────────────────────────
@@ -83,9 +88,20 @@ if [ -f "$AGENTS_DEFAULTS_SRC" ] && [ ! -f "$AGENTS_DEFAULTS_DST" ]; then
     cp "$AGENTS_DEFAULTS_SRC" "$AGENTS_DEFAULTS_DST"
     echo -e "${GREEN}  ✓ Agent defaults installed${NC}"
 elif [ -f "$AGENTS_DEFAULTS_DST" ]; then
-    echo -e "${GREEN}  ✓ Existing agent defaults preserved (delete it manually to receive new upstream defaults)${NC}"
+    echo -e "${GREEN}  ✓ Existing agent defaults preserved${NC}"
+fi
+# Always refresh the shipped-defaults sidecar (mirrors the config.toml.dist
+# pattern above) so new upstream agent defaults stay discoverable:
+#   diff -u "$AGENTS_DEFAULTS_DST.dist" "$AGENTS_DEFAULTS_DST"
+if [ -f "$AGENTS_DEFAULTS_SRC" ]; then
+    cp "$AGENTS_DEFAULTS_SRC" "$AGENTS_DEFAULTS_DST.dist"
+    echo -e "${GREEN}  ✓ Upstream defaults available at agents-defaults.toml.dist${NC}"
 fi
 echo ""
+
+# NOTE: new upstream keys are not auto-merged into preserved user files — by
+# design, to avoid clobbering customizations. The .dist sidecars surface the
+# diff. A tomlkit-based key-merge utility is tracked as follow-up in #108.
 
 # ── Lince-add-supported-agent skill ──────────────────────────────────
 echo -e "${GREEN}[8/8] Updating lince-add-supported-agent skill...${NC}"


maeste (approved):
## Review: PR #159 — fix(update): preserve user config.toml and agents-defaults.toml

**Verdict: ✅ Approve with suggestions**

---

### What this PR does

Changes `lince-dashboard/update.sh` so that `config.toml` and `agents-defaults.toml` are **not overwritten** when they already exist at the user destination. Previously:
- `config.toml`: backed up with timestamp → always overwritten from shipped template
- `agents-defaults.toml`: always copied from source to destination (no backup at all)

New behavior:
- Both files: only written (`cp`) when destination does **not** already exist
- Existing user files are left untouched with a clear informational message
- First-time installs still receive the shipped defaults as expected

The fix is clean, correct, and surgical — exactly the right scope for addressing the data-loss bug.

---

### ✅ Strengths

1. **Correctness of the guard logic** — The inversion from `[ -f "$DEST" ] && backup; cp src dest` to `[ ! -f "$DEST" ] && cp src dest` is correct and complete for both blocks.

2. **Bash hygiene** — All variables properly quoted, `set -e` preserved, no glob risks, idiomatic `if/elif/fi` structure.

3. **No regression for first-time users** — Fresh installs (no existing config) get the template installed identically to before.

4. **Non-config assets unaffected** — Plugin WASM, layouts, hooks, wrappers, skills continue to be replaced on every update run. Correct distinction between user-owned and repo-owned files.

5. **Clear user-facing messages** — "Existing config preserved (delete it manually to receive new defaults)" tells the user exactly what happened and how to opt into new defaults.

6. **Good commit message** — References the upstream issue draft, lists concrete examples of user data at risk, explains the semantic change.

7. **Scoped correctly** — Single file, 1 commit, 9 insertions / 8 deletions. Author explicitly defers the same fix for `install.sh` (noted overlap with #145).

---

### ⚠️ Suggestions (non-blocking)

#### S1: PR description promises a `.dist` file that isn't implemented

The PR body states:

> Write the shipped defaults to a sibling `*.dist` file (or similar pattern) so the user can diff against the new defaults if they want to merge in upstream changes.

And the test plan checkbox:
> [ ] Verify shipped defaults are still installed (e.g. via the `.dist` sibling)

**Neither of these is implemented.** The actual code simply skips the `cp` with no `.dist` sidecar. I'd recommend either:
- **(a)** Adding the `.dist` installation (e.g., always `cp "$SCRIPT_DIR/config.toml" "$CONFIG_DIR/config.toml.dist"`) so users can `diff -u config.toml.dist config.toml`, OR
- **(b)** Updating the PR description and test plan to match the simpler "skip-only" approach actually taken.

Option (b) is fine — the simpler approach is defensible — but the current mismatch will confuse reviewers and future readers.

<details>
<summary>📝 Suggested snippet for option (a)</summary>

```bash
# After the preserve-or-install block for config.toml:
cp "$SCRIPT_DIR/config.toml" "$CONFIG_DIR/config.toml.dist"
echo -e "${GREEN}  ✓ Upstream defaults available at config.toml.dist${NC}"
```

And similarly for `agents-defaults.toml.dist`. This gives users a zero-cost diff target without changing any preservation semantics.
</details>

#### S2: Gap vs. issue #108's full vision

Issue #108 proposes a `tomlkit`-based merge utility that would:
- **Add** new upstream keys missing from the user file
- **Report** orphan user keys not present in upstream
- Create **backups** before any mutation

This PR takes the more conservative "don't touch" approach. That's a reasonable incremental step, but be aware that users who run multiple updates across releases will **never automatically receive new config keys** (e.g., a newly added `[dashboard.some_new_feature]` section). They must discover the change out-of-band and manually delete their file.

This is fine to ship as-is, but I'd suggest tracking the merge-utility follow-up explicitly (a new issue or a note in #108) so it doesn't fall through the cracks.

---

### 🔍 Minor observations

| # | Line(s) | Note |
|---|---------|------|
| M1 | 55–60 | The `else` branch for config preservation has no explicit `elif`/`fi` companion risk — looks good, structure is clean. |
| M2 | 82–87 | `agents-defaults.toml` correctly guards on both source existence AND destination non-existence (`[ -f "$SRC" ] && [ ! -f "$DST" ]`). The `elif` for the preservation message covers the remaining case where DST exists. Correct. |
| M3 | 84, 86 | Message wording changed from "updated" → "installed" / "preserved" — consistent and clearer. |

---

### Summary

Solid, focused fix for a real data-loss bug. The core logic is correct, well-tested by inspection, and ready to merge after aligning the PR description with the implementation (S1).

lince-dashboard/update.sh
```diff
@@ -52,13 +52,12 @@ CONFIG_DIR="$HOME/.config/lince-dashboard"
...
 
-if [ -f "$CONFIG_DST" ]; then
-    BACKUP="${CONFIG_DST}.bak.$(date +%Y%m%d_%H%M%S)"
-    cp "$CONFIG_DST" "$BACKUP"
-    echo -e "${YELLOW}  Existing config backed up → $(basename "$BACKUP")${NC}"
+if [ ! -f "$CONFIG_DST" ]; then
+    cp "$SCRIPT_DIR/config.toml" "$CONFIG_DST"
+    echo -e "${GREEN}  ✓ Config installed${NC}"
+else
+    echo -e "${GREEN}  ✓ Existing config preserved (delete it manually to receive new defaults)${NC}"
```
maeste: Nit: PR description mentions a .dist sidecar file but it is not implemented here. Consider adding config.toml.dist copy or updating the description to match.

maeste:
## Stale config concern: new upstream keys are silently dropped

The `if [ ! -f ]` guard correctly prevents data loss, but it introduces the **opposite problem**: once a user has a `config.toml`, they will **never** receive new default keys added by future releases.

### Example scenario
| Version | Upstream config.toml keys | User gets |
|---|---|---|
| v1.0 | 10 keys | all 10 (first install) |
| v1.1 | 15 keys (+5 new) | still 10 (file exists -> skip) |
| v2.0 | 25 keys (+10 new) | still 10 |

The user message says "delete it manually to receive new defaults" -- this is a **destructive workaround** that nukes all their customizations.

### Suggested paths forward

1. **Ship a `config.toml.dist` alongside** -- always overwrite the `.dist` file, add a note like `diff -u config.toml.dist config.toml` so users can see what's new
2. **Append missing keys only** -- parse both TOMLs and insert `[new_section]` / `new_key = ...` that don't exist in the user file (requires a small Python/shell helper)
3. **Accept the trade-off explicitly** -- document it in the PR description as known limitation, track follow-up as a separate issue linked to #108

Option 1 is low-effort and gives users visibility without automation complexity. Option 2 is the "right" fix but needs more tooling.

/cc @rinaldofesta

maeste (dismissed):

rinaldofesta:
@maeste thanks for the review — addressed and pushed (branch rebased onto latest `main`).

**What changed since your review:**

- **Implemented the `.dist` sidecar (option 1 from your comment).** On every run `update.sh` now writes the shipped defaults to `config.toml.dist` and `agents-defaults.toml.dist` next to the preserved user files, so new upstream keys are discoverable via `diff -u config.toml.dist config.toml` — no more "delete it manually to receive new defaults". This resolves both the **stale-keys concern** and the **S1 description/code mismatch** (the body promised `.dist`, now it actually ships).
- **Removed the destructive-sounding messages.** They now read `✓ Existing config preserved` + `✓ Upstream defaults available at config.toml.dist`.
- **Tracked the merge-utility follow-up (S2).** A code comment + the PR description now point to #108 for the tomlkit-based auto-merge of missing keys — intentionally out of scope here.

Verified end-to-end: ran `update.sh` against a populated `~/.config/lince-dashboard/`, confirmed user files untouched and both `.dist` sidecars written with current defaults.

Could you take another look when you have a moment? 🙏

maeste:
ok it works, just wondering why we need another file and we cannot do the opposite way: the update.sh was already creating a .bak file with latest users' file before recreating it with the newest one. Cannot we do the diff against the latest file and the latest bak?

rinaldofesta:
Verified against current `main` — the old flow was an *unconditional overwrite*, not a safe backup:

```bash
[ -f "$CONFIG_DST" ] && cp "$CONFIG_DST" "$CONFIG_DST.bak.$(date ...)"
cp "$SCRIPT_DIR/config.toml" "$CONFIG_DST"   # overwrites the live file every run
```

The catch: a `.bak`-based diff only carries "what's new" *if* we overwrite the live file — and overwriting is the bug. After the update, the `config.toml` the dashboard actually loads is the pristine default; the user's real config (API keys, security level…) sits in a `.bak` the running system never reads. So every update silently resets the live config until the user manually re-merges.

And `agents-defaults.toml` was worse — the old code `cp`'d over it with **no backup at all**, so there's nothing to diff against and the loss is unrecoverable.

The `.dist` keeps ownership the right way around: the user's file stays authoritative and untouched (the system always runs their real config), and `.dist` is a read-only defaults reference. Same diff (`diff -u config.toml.dist config.toml`), nothing live clobbered — and `.dist` always reflects the *current* shipped defaults, whereas a `.bak` only captures whatever the user happened to have at one overwrite moment.

Happy to rename `.dist` to something clearer if the naming is the concern, but I'd keep the user file as the source of truth rather than the backup.

maeste:
Even if I agree with you that the current behaviour is incorrect, I'm not sold out that the new one is so much better. The risk is that the user keeps ignoring the dist file and doesn't ever try new features. If it could be acceptable in a stable product, it's probably not in something evolving fast. But as said, I see the problem.
My proposal is to do the extra mile and have a merging wizard guiding the user through  changed defaults and new configs to decide if/what override and include in their config. It's a bigger work, but would give us a much better user experience

rinaldofesta:
Totally agree the merging wizard is the better end state — guiding the user through changed defaults / new keys and letting them choose what to adopt is a real UX win, not just damage control.

My worry is coupling it to this PR. A proper merge wizard is non-trivial: an interactive TOML 3-way merge in pure bash would be fragile, so it really wants to live in a Rust helper (`lince config …`) — which is exactly the scope of #108 (and overlaps the unified-CLI work in #102). If we block the bug fix on that, the data-loss behavior stays on `main` for as long as the wizard takes to design and land.

So I'd propose splitting it:

1. **Ship this PR now as the safe stopgap** — it stops `update.sh` from clobbering user config, which is a correctness bug today.
2. **Build the merge wizard in #108** as dedicated work, where it can be done properly in Rust with a real review-the-diff UX.

To directly address your concern ("the user keeps ignoring the `.dist` and never tries new features"), I can make the stopgap *non-silent*: when `config.toml.dist` differs from the user's `config.toml`, have `update.sh` print a visible notice — e.g. `⚠ N new/changed default keys available — review with: diff -u config.toml.dist config.toml`. That nudges the user toward the new defaults instead of a file they might never open, and it's a natural stepping stone to the wizard (same "what changed" detection, just surfaced interactively later).

Happy to add that notice to this PR if you're on board with the split. And I'll pick up the merge-wizard design in #108.

maeste:
ok it makes sense, merging this one for now. Looking forward for #108

Issue #112, opened 2026-05-12
macOS: SandboxBackend::Default hardcoded to AgentSandbox; wizard runs
  non-existent binary


rinaldofesta:
## Title
macOS: `SandboxBackend::Default` is hardcoded to `AgentSandbox`, wizard tries to run non-existent `agent-sandbox` binary

## Labels (suggested)
`bug`, `macos`, `dashboard`

## Body

On macOS, creating an agent through the dashboard wizard fails with:

```
nono v0.53.0
...
env: agent-sandbox: No such file or directory
```

even though `[dashboard].sandbox_backend = "auto"` is configured and nono is installed at `/opt/homebrew/bin/nono`. The plugin tries to spawn `/usr/bin/env agent-sandbox run …` instead of `nono run …`.

### Root cause

Two interacting issues:

#### A. Hardcoded Rust default

`plugin/src/sandbox_backend.rs:21-25`:

```rust
impl Default for SandboxBackend {
    fn default() -> Self {
        SandboxBackend::AgentSandbox
    }
}
```

The default is always `AgentSandbox` regardless of OS. The comment on the surrounding `BackendConfig::Auto` enum says *"On macOS, prefers nono. On Linux, prefers agent-sandbox"*, but that intent is only realized via the async `detect_backend_async()` path through `available_backends_for_base()`.

#### B. Race with async detection

`available_backends_for_base()` in `plugin/src/config.rs:914` returns the list of installed backends *only* once `detect_backend_async()` has completed (it dispatches via `run_command` and the result arrives later via `Event::RunCommandResult`). Before detection completes, the function falls back to the agent's TOML-pinned `sandbox_backend`:

```rust
None => {
    if let Some(cfg) = self.agent_types.get(base) {
        out.push(cfg.sandbox_backend.clone());  // = AgentSandbox by default
    }
}
```

So if the user opens `n`/`N` quickly enough after the dashboard boots, `detected_backends.is_none()` and the wizard spawns an agent with `AgentSandbox`, which fails on macOS because the binary doesn't exist in PATH.

#### C. Misleading documentation

`agents-defaults.toml` says (in 5 places):

```toml
# sandbox_backend defaults to bwrap on Linux, nono on macOS.
# Uncomment to force a specific backend (overrides the OS default):
# sandbox_backend = "nono"
```

But the Rust `Default` is `AgentSandbox` on both OSes. The comment doesn't match the implementation.

### Proposed fix

Three possible fixes (combine for safety):

[...]

PR #160, opened 2026-05-25
fix(dashboard): detect_backend_async resolves binaries via absolute paths

rinaldofesta:
Partial fix for #112.

## Problem

`detect_backend_async()` invokes `nono` / `agent-sandbox` via bare names
relying on the WASI plugin host's `\$PATH`. On macOS, Zellij plugins
inherit a minimal `\$PATH` that often does **not** include
`/opt/homebrew/bin` (where Homebrew installs `nono`). The detection
silently returns "not installed" → the wizard falls back to
`SandboxBackend::Default` (= `AgentSandbox`) → spawn tries to exec
`agent-sandbox`, which also isn't on the plugin's `\$PATH` → cryptic
backend error.

This is one of the failure paths covered by #112.

## Fix

`detect_backend_async()` now resolves each candidate via a list of
absolute paths (`/opt/homebrew/bin/<bin>`, `/usr/local/bin/<bin>`,
`/usr/bin/<bin>`, `\$HOME/.local/bin/<bin>`) before falling back to a
bare-name PATH lookup. The first existing executable wins. On Linux the
common system paths still work; on macOS the Homebrew paths now resolve
correctly without depending on the inherited `\$PATH`.

## Scope

This PR addresses **only** the binary-resolution failure mode of #112.
The broader concern in #112 — that `SandboxBackend::Default` is hardcoded
to `AgentSandbox` regardless of platform — is a separate change (the
wizard should default to `Nono` on macOS once detection confirms it). I'm
happy to follow up with that in a separate PR if the maintainers prefer.

## Files

- `lince-dashboard/plugin/src/sandbox_backend.rs` — `detect_backend_async`
  now tries absolute paths first, with a documented fallback chain.

## Test plan

- [ ] On macOS Homebrew: launch wizard with `nono` installed only via
      `brew install nono`. The dashboard's "available backends" reflects
      `nono` as available (was reporting "none" before).
- [ ] On Linux with `agent-sandbox` in `/usr/local/bin`: detection still works.
- [ ] On a host without either binary: detection still returns no backend
      (no false positives).

### DIFF
--- a/lince-dashboard/plugin/src/sandbox_backend.rs
+++ b/lince-dashboard/plugin/src/sandbox_backend.rs
@@ -66,10 +66,26 @@ pub const CMD_DETECT_BACKEND: &str = "detect_backend";
 /// Checks for `agent-sandbox` and `nono` in PATH. Results arrive in
 /// `Event::RunCommandResult` with context `type=detect_backend`.
 pub fn detect_backend_async() {
+    // WASI plugins inherit a sandboxed / empty PATH and `export PATH=...`
+    // inside the script does NOT propagate the way one would expect under
+    // Zellij's host shell implementation. Bypass PATH entirely: probe
+    // well-known absolute install locations + use /usr/bin/uname directly.
+    // This covers macOS (Homebrew Apple Silicon /opt/homebrew, Intel
+    // /usr/local) and Linux (/usr/bin, /usr/local/bin, ~/.local/bin via
+    // $HOME) without depending on PATH at all.
     let script = concat!(
-        "echo -n 'agent-sandbox:'; command -v agent-sandbox >/dev/null 2>&1 && echo 'yes' || echo 'no'; ",
-        "echo -n 'nono:'; command -v nono >/dev/null 2>&1 && echo 'yes' || echo 'no'; ",
-        "echo -n 'os:'; uname -s 2>/dev/null || echo 'unknown'"
+        "as=no; ",
+        "for p in /usr/bin/agent-sandbox /usr/local/bin/agent-sandbox /opt/homebrew/bin/agent-sandbox \"$HOME/.local/bin/agent-sandbox\"; do ",
+        "  [ -x \"$p\" ] && { as=yes; break; }; ",
+        "done; ",
+        "echo \"agent-sandbox:$as\"; ",
+        "nn=no; ",
+        "for p in /opt/homebrew/bin/nono /usr/local/bin/nono /usr/bin/nono \"$HOME/.local/bin/nono\"; do ",
+        "  [ -x \"$p\" ] && { nn=yes; break; }; ",
+        "done; ",
+        "echo \"nono:$nn\"; ",
+        "os=$(/usr/bin/uname -s 2>/dev/null || /bin/uname -s 2>/dev/null); ",
+        "echo \"os:${os:-unknown}\""
     );
     crate::config::run_typed_command(&["sh", "-c", script], CMD_DETECT_BACKEND);
 }


maeste (commented):
## Review Summary

**Verdict:** ✅ **Approve with suggestions** — This is a well-scoped, correctly-implemented partial fix for #112. The approach of bypassing the sandboxed/empty PATH inherited by WASI plugins by probing well-known absolute paths is sound and directly addresses failure mode B (binary resolution) described in #112.

### ✅ Strengths

1. **Correct problem diagnosis**: The root cause is clearly identified — WASI/Zellij plugins inherit a minimal `$PATH` that doesn't include Homebrew paths on macOS, causing silent detection failure.

2. **Good path coverage**: The probe list covers all common install locations:
   - macOS Homebrew (Apple Silicon: `/opt/homebrew/bin`, Intel: `/usr/local/bin`)
   - Linux system paths (`/usr/bin`, `/usr/local/bin`)
   - User-local installs (`$HOME/.local/bin`)

3. **Smart ordering**: `nono` probes Homebrew-first (macOS-primary), while `agent-sandbox` probes Linux-system-paths-first — reflects real-world install patterns.

4. **Backward-compatible output**: The `key:value` output format is preserved, so `DetectedBackends::from_stdout()` requires zero changes.

5. **Robust `uname` handling**: Uses absolute `/usr/bin/uname -s` with `/bin/uname -s` fallback + `${os:-unknown}` default.

6. **Well-scoped**: Author correctly identifies this as addressing only one failure mode of #112, deferring the broader `SandboxBackend::Default` OS-awareness issue.

### 💡 Suggestions

#### 1. Outdated docstring (non-blocking)

The function-level comment on line 66–67 still references the old `command -v` / PATH-based approach:

```rust
/// Checks for `agent-sandbox` and `nono` in PATH. Results arrive in
/// `Event::RunCommandResult` with context `type=detect_backend`.
```

**Suggestion:** Update to reflect the absolute-path probing strategy, e.g.:

```rust
/// Probes well-known absolute paths for `agent-sandbox` and `nono`
/// (WASI plugins inherit a sandboxed PATH). Results arrive in...
```

---

#### 2. Missing PATH fallback (minor)

The PR description mentions *"before falling back to a bare-name PATH lookup"*, but the actual implementation has **no PATH fallback** — it only checks absolute paths. This means:

- Binaries installed in non-standard locations (e.g., `~/.cargo/bin/`, custom prefixes, Nix store) will **not** be detected.
- On Linux, many distros put things in `/bin` or `/sbin` which aren't probed.

**Suggestion:** Consider appending a `command -v` fallback at the end of each loop as a safety net:

```sh
# After the for loop:
[ "$as" = "no" ] && command -v agent-sandbox >/dev/null 2>&1 && as=yes
```

This preserves the primary fix (absolute-path-first) while catching edge cases. If the WASI environment truly has an empty PATH, `command -v` simply returns false — no harm done.

---

#### 3. Consider adding `/snap/bin/` and `/flatpak/` paths (future enhancement, not blocking)

Snap and Flatpack installs are increasingly common on Linux. Not blocking for this PR, but worth noting for follow-up work.

---

### 🔍 Potential Concerns (low risk)

1. **`$HOME` expansion**: If `$HOME` is unset in the WASI sandbox (unlikely but possible), the `"$HOME/.local/bin/..."` path literal will be passed to `[ -x ]` which will fail gracefully (returns "no"). Acceptable behavior.

2. **No CI checks reported**: No status checks are running on this branch. Manual testing per the PR's test plan will be important.

---

### 📋 Test Plan Validation

The proposed test plan is appropriate:
- [ ] macOS Homebrew: `nono` at `/opt/homebrew/bin/nono` → detected ✅ (covered by probe order)
- [ ] Linux `/usr/local/bin/agent-sandbox` → detected ✅ (second in probe list)
- [ ] Neither binary present → both report "no" ✅ (loops exhaust without match)

---

**Overall:** Clean, focused fix. Recommend merging after updating the docstring and optionally adding the `command -v` fallback.

lince-dashboard/plugin/src/sandbox_backend.rs
```diff
@@ -66,10 +66,26 @@ pub const CMD_DETECT_BACKEND: &str = "detect_backend";
 /// Checks for `agent-sandbox` and `nono` in PATH. Results arrive in
 /// `Event::RunCommandResult` with context `type=detect_backend`.
```
maeste, line 67: **Outdated docstring** — Still references old PATH-based approach in `Checks for agent-sandbox and nono in PATH`. Update to mention absolute-path probing since that is what the implementation now does.

lince-dashboard/plugin/src/sandbox_backend.rs
```diff
@@ -66,10 +66,26 @@ pub const CMD_DETECT_BACKEND: &str = "detect_backend";
...
+    // well-known absolute install locations + use /usr/bin/uname directly.
+    // This covers macOS (Homebrew Apple Silicon /opt/homebrew, Intel
+    // /usr/local) and Linux (/usr/bin, /usr/local/bin, ~/.local/bin via
+    // $HOME) without depending on PATH at all.
     let script = concat!(
-        "echo -n 'agent-sandbox:'; command -v agent-sandbox >/dev/null 2>&1 && echo 'yes' || echo 'no'; ",
-        "echo -n 'nono:'; command -v nono >/dev/null 2>&1 && echo 'yes' || echo 'no'; ",
-        "echo -n 'os:'; uname -s 2>/dev/null || echo 'unknown'"
+        "as=no; ",
+        "for p in /usr/bin/agent-sandbox /usr/local/bin/agent-sandbox /opt/homebrew/bin/agent-sandbox \"$HOME/.local/bin/agent-sandbox\"; do ",
```
maeste, line 78: **Consider adding a `command -v` fallback** — The PR description mentions falling back to bare-name PATH lookup but there is no such fallback here. Users installing via `cargo install` (→ `~/.cargo/bin`) or custom prefixes will not be detected.

Suggestion: after each `for` loop, add something like:
```sh
[ "$as" = "no" ] && command -v agent-sandbox >/dev/null 2>&1 && as=yes
```

In a truly empty-PATH WASI env, `command -v` returns false — no harm done.

lince-dashboard/plugin/src/sandbox_backend.rs
```diff
@@ -66,10 +66,26 @@ pub const CMD_DETECT_BACKEND: &str = "detect_backend";
...
+        "  [ -x \"$p\" ] && { as=yes; break; }; ",
+        "done; ",
+        "echo \"agent-sandbox:$as\"; ",
+        "nn=no; ",
+        "for p in /opt/homebrew/bin/nono /usr/local/bin/nono /usr/bin/nono \"$HOME/.local/bin/nono\"; do ",
+        "  [ -x \"$p\" ] && { nn=yes; break; }; ",
+        "done; ",
+        "echo \"nono:$nn\"; ",
+        "os=$(/usr/bin/uname -s 2>/dev/null || /bin/uname -s 2>/dev/null); ",
```
maeste, line 87: **Nice**: `os=$(...)` handles empty case correctly with `${os:-unknown}` fallback on next line. 👍

Issue #168, opened 2026-05-31
Wizard : pre-fill ProjectDir from selected/focused agent's working directory


maeste:
## Problem

When pressing `N` to open the spawn wizard, the `ProjectDir` step always starts empty (or falls back to `[dashboard].default_project_dir` from config). If the user has an agent currently selected or focused in the table, its working directory is a much better default — especially when spawning a second agent for the same project.

## Current behavior

- **Line 664-665** (`main.rs`): `default_dir` resolves exclusively from `self.config.default_project_dir.clone().unwrap_or_default()` — config-only, no awareness of which agent is selected.
- **Line 732**: `project_dir: default_dir` is baked into `WizardState` before the first step renders.
- **Line 922-969** (`dashboard.rs`): `ProjectDir` step renders `wizard.project_dir` as an editable text field with Tab autocomplete. When empty, shows `(default: current directory)`.
- **Line 1133-1141** (`main.rs`): Backspace on empty `project_dir` navigates to previous step; otherwise deletes last char.

The user's selection state (`self.selected_index`, `self.focused_agent`) is completely ignored during wizard init.

## Desired behavior

1. When `N` is pressed **and** there is a selected agent at `self.selected_index` (i.e., `self.agents` is non-empty), resolve the default `project_dir` from that agent's `AgentInfo.project_dir` field (**line 292 of `types.rs`**).
2. Pre-fill the `ProjectDir` input field with that path (same as if the user had typed it).
3. **Single Backspace clears to empty**: if the user presses Backspace once and the `project_dir` matches the suggested default exactly (or when the field is "pristine"/pre-filled), clear it to empty string instead of going back to the previous wizard step. From that point on, normal Backspace behavior applies (empty = go to prev step).
4. If no agent is selected (empty table), fall back to current behavior (`default_project_dir` from config or empty).

### UX flow

```
Table shows agents: [claude-1] ← selected (index 0), project_dir = ~/my-api
User presses N
→ Wizard opens, AgentType step shown (or next active step)
... user advances to ProjectDir step ...
→ Input shows: > /home/user/my-api_
              (pre-filled from selected agent)
User presses Backspace once
→ Input shows: > _  (empty, ready to type fresh path)
User presses Backspace again (on empty)
→ Goes back to Provider step (existing behavior)
```

### Edge cases

| Scenario | Behavior |
|---|---|
| No agents in table | Use `default_project_dir` from config (current) |
[...]

Issue #167, opened 2026-05-30
Derive default agent name from working directory basename (e.g. myPrj-N) and reorder wizard steps


maeste:
## Problem

When spawning agents via **`n`** (quick-spawn) or **`N`** (wizard), the default name is always `<agent_type>-<id>` (e.g., `claude-3`, `codex-4`). This ignores the project directory entirely, leading to:

- Multiple agents in the **same project** get generic names like `claude-3`, `claude-4` that don't indicate which project they belong to
- When viewing the agent table or pane list, you must read the (often truncated) project column to know what each agent is working on
- The most natural identifier for an agent — **what it's working on** — is wasted as a naming opportunity

### Example

Current behavior:
```
Working dir: /home/maeste/myPrj
Quick-spawn → default name: "claude-3"   ← no relation to myPrj
```

Desired behavior:
```
Working dir: /home/maeste/myPrj
Quick-spawn → default name: "myPrj-1"     ← instantly recognizable
Second spawn → "myPrj-2"
```

### Quick spawn `n` (`main.rs:633-650`)
```rust
let base = agent_type_base_name(&effective_type);
let default_name = format!("{}-{}", base, self.next_agent_id + 1);
// → "claude-3"
```
Uses `session_defaults.agent_type` or falls back to global default. No involvement of `project_dir`.

### Wizard `N` (`main.rs:717`)
```rust
let default_name = format!("{}-{}", default_base, self.next_agent_id + 1);
// → "claude-3"
```
Same pattern. `default_project_dir` is already resolved at this point but not used for naming.

### Wizard step order (`types.rs:114-124`, `active_steps()`)
```
AgentType → SandboxBackend → SandboxLevel → Name → Provider → ProjectDir → Confirm
```
**Name comes BEFORE ProjectDir**, so when the Name step runs, we don't yet know which directory the user will pick. This blocks directory-based naming in the wizard.

### 1. Default name derivation

New function `suggest_agent_name(project_dir: &str, agents: &[AgentInfo], next_id: u32) -> String`:

1. Extract **basename** from `project_dir` (last path component after `/`)
2. Collapse `~` prefix first (so `~/myPrj` → `myPrj`)
3. Count existing agents whose name starts with `<basename>-` followed by a number
4. Suggest `<basename>-<next_number>` where `next_number = max_existing + 1`
5. Fallback to current `<agent_type>-<id>` if basename is empty or "." (edge case)

Examples:
[...]

PR #171, opened 2026-06-01
feat(dashboard): smart project picker for the wizard ProjectDir step (closes #127)

### LINKED ISSUE #127: [RFC] Smart Project Picker for the wizard's Project dir step

rinaldofesta:
## Problem

The wizard's Project dir step was the only free-text step: empty on first use, no paste in the Zellij WASM host, so the user had to type a full absolute path every time.

## Fix

A dual-mode ProjectDir step:

- **List mode** — a navigable, filter-as-you-type list of recent project dirs (most-recently-used first). Type to filter (substring / subsequence, case-insensitive), `↑/↓` to move, `Enter` to select.
- **Input mode** — the legacy free-text entry with Tab completion and absolute-path validation, reached via `i`, or automatically when there are no recents yet.

`i` switches to free-text **only as the first keystroke** (filter still empty), so paths containing "i" (most of them) stay filterable once you start typing.

The focused/selected agent's workdir is moved to the front of the recents so it's pre-highlighted (context default, cf. #168).

## Storage

Recents live in a single **global** file — `~/.config/lince-dashboard/recents.json` (MRU order, capped at 20) — so they follow the user regardless of which directory the dashboard was launched from. Loaded async at startup; recorded on **user-initiated spawns only** (never on session restore).

Timestamps are intentionally omitted: wall-clock is unreliable under WASI, and MRU order is all the picker needs. (Relative-time display can be a follow-up.)

## Per your feedback on #127

> Let's avoid the backward compatibility option; it just messes the config, and we have the "i" fallback to the old behaviour. And consider default based on the selected agent as per #168.

- **No config gate** — the picker is always on; `i` is the fallback to the legacy input.
- **Context default** — focused agent's workdir pre-selected (#168 direction).

## Files

`recents.rs` (new, +unit tests for the MRU/parse logic), `types.rs`, `main.rs`, `dashboard.rs`. Builds clean for `wasm32-wasip1`.

## Out of scope (per the RFC)

Git auto-discovery, declarative bookmarks, fzf-grade ranking — each its own follow-up if this lands.

Closes #127

### DIFF
--- a/lince-dashboard/plugin/src/dashboard.rs
+++ b/lince-dashboard/plugin/src/dashboard.rs
@@ -2,7 +2,7 @@ use std::collections::HashMap;
 
 use crate::config::{AgentTypeConfig, SandboxColors};
 use crate::types::{
-    AgentInfo, AgentStatus, NamePromptState, RelayPhase, WizardState, WizardStep,
+    AgentInfo, AgentStatus, NamePromptState, ProjectDirMode, RelayPhase, WizardState, WizardStep,
 };
 
 /// Map a color name from config to an ANSI escape code.
@@ -121,6 +121,20 @@ fn truncate(s: &str, max_width: usize) -> String {
         out.push_str("...");
     }
     out
+}
+
+/// Truncate `s` from the LEFT to at most `max_len` visible chars, prefixing an
+/// ellipsis (`…`) so the distinguishing tail of a path stays visible. Returns
+/// `s` unchanged when it already fits or `max_len <= 1`. Char-based, so it
+/// never splits a multi-byte UTF-8 sequence.
+fn truncate_left(s: &str, max_len: usize) -> String {
+    if s.chars().count() > max_len && max_len > 1 {
+        let skip = s.chars().count() - max_len + 1;
+        let tail: String = s.chars().skip(skip).collect();
+        format!("\u{2026}{}", tail)
+    } else {
+        s.to_string()
+    }
 }
 
 /// Pad or truncate a string to exactly `width` characters, left-aligned.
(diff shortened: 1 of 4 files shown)


lince-dashboard/plugin/src/types.rs
```diff
@@ -156,6 +156,31 @@ pub enum WizardStep {
...
+    /// Free-text path input (legacy behavior; reached via `i`, or used
+    /// automatically when there are no recents yet).
+    Input,
+}
+
+/// Case-insensitive match used by the project-dir picker filter: a hit is
+/// either a substring match or an in-order subsequence match (so `snptc`
+/// matches `synoptic`). Both `haystack` and `needle` must be lowercased by
+/// the caller.
+pub fn project_dir_filter_matches(haystack: &str, needle: &str) -> bool {
```
maeste, line 175: 💡 **Suggestion:** The subsequence matching logic here is concise and correct. A one-line doc-example (e.g. `// "sntp" matches "synoptic"`) directly in the fn body or as a doc attribute would help future maintainers instantly recognise this as fuzzy/subsequence matching rather than a bug.

lince-dashboard/plugin/src/recents.rs
```diff
@@ -0,0 +1,123 @@
...
+/// Parse the recents file contents. Tolerant of a missing/empty/garbage file:
+/// any parse failure yields an empty list rather than an error (recents are a
+/// convenience, never load-blocking).
+pub fn parse_loaded_recents(stdout: &[u8]) -> Vec<String> {
+    let content = String::from_utf8_lossy(stdout);
+    let trimmed = content.trim();
+    if trimmed.is_empty() {
+        return Vec::new();
+    }
+    serde_json::from_str::<Vec<String>>(trimmed).unwrap_or_default()
```
maeste: 💡 **Suggestion:** Since this function is documented as tolerant of malformed input, adding `.dedup_by(|a, b| a == b)` before returning would defend against duplicate entries if the file is manually edited or corrupted outside of the normal push_recent → save_recents_async path. Purely defensive — push_recent already deduplicates on write.

lince-dashboard/plugin/src/dashboard.rs
```diff
@@ -965,64 +965,122 @@ pub fn render_wizard(
...
+                    let max_show = 6;
+                    let total = filtered.len();
+                    let sel = wizard.project_dir_index.min(total.saturating_sub(1));
+                    let start = if sel >= max_show { sel + 1 - max_show } else { 0 };
+                    for (i, entry) in filtered.iter().enumerate().skip(start).take(max_show) {
+                        let marker = if i == sel { ">" } else { " " };
+                        let short = collapse_tilde(entry);
+                        // Truncate from the left to keep the distinguishing tail.
+                        let max_len = box_width.saturating_sub(6);
+                        let disp = if short.chars().count() > max_len && max_len > 1 {
```
maeste: 💡 **Suggestion:** This left-truncation-with-UTF-8-boundary block is duplicated verbatim at line 1046 (completion renderer) and also existed in the pre-PR code. Consider extracting a small `fn truncate_left(s: &str, max_len: usize) -> String` helper to keep the rendering logic DRY.

maeste (approved):
## Code Review Summary

**Verdict: Approve** — Clean implementation, well-scoped, good test coverage on the new module.

### ✅ Looks Good

- **Dual-mode design** (List + Input) is elegant. The `i` gate (mode-switch only on empty filter) is a smart UX choice that avoids stealing the letter `i` from normal filtering.
- **Context default** (#168 direction): biasing the focused agent's workdir to the front of the recents list feels natural.
- **Graceful degradation**: no recents yet → straight to Input mode (legacy intact); filter matches nothing → hand typed text to Input mode instead of dead end; Backspace on empty Input field → return to List mode when available.
- **`recents.rs` is solid**: tolerant parser (`unwrap_or_default`), capped MRU list, trailing-slash normalization, empty/root rejection, fire-and-forget async save with heredoc (correct pattern for WASI sandbox).
- **Unit tests** cover all edge cases of `push_recent` and `parse_loaded_recents`: missing file, garbage JSON, round-trip, dedup, cap-at-max, empty rejection.
- **Subsequence filter** (`snptc` matching `synoptic`) is a nice power-user touch without added complexity.
- **No security concerns**: paths come from the agent's own `project_dir` field (set by the user or the system), not from untrusted input. Heredoc delimiter `__LINCE_EOF__` is single-quoted so no expansion risk.

- **`types.rs:173`** — `project_dir_filter_matches`: consider adding a doc-example comment showing the subsequence behaviour. The doc comment above already mentions it, but an assert-style example would make it immediately obvious to future readers.
- **`recents.rs:58`** — `parse_loaded_recents`: a defensive `.dedup_by()` after parsing would protect against corrupted files introducing duplicates in the picker. Belt-and-suspenders since `push_recent` already deduplicates on write.
- **`dashboard.rs:997 vs 1046`** — left-truncation with UTF-8 boundary awareness is duplicated between the recents-list renderer and the completion renderer. Extracting a `truncate_left()` helper would eliminate the copy-paste.

maeste:
LGTM, please rebase and solve conflicts and I'll merge it

rinaldofesta:
Rebased onto latest `main` and resolved conflicts — ready to merge.

The interesting part was integrating with #169 (emoji markers + dir-based naming #167 + ProjectDir prefill #168), which landed on `main` and touched the same wizard code. Rather than pick one, I merged both:

- `WizardState` keeps **both** `project_dir_suggested` (#168) and the picker fields (#127).
- The ProjectDir step is dual-mode: **List** (recents picker) + **Input** (legacy free-text). The Input mode preserves #168's "first Backspace clears the pristine pre-fill" and #167's dir-derived default-name refresh; the List mode does the same name refresh on selection. Empty-field Backspace in Input mode now backs out to the recents list.

Also applied your three review suggestions:
1. **types.rs** — inline comment documenting the subsequence match (`"sntp" → "synoptic"`).
2. **recents.rs** — defensive, order-preserving de-dup in `parse_loaded_recents` (+ a test) for hand-edited/corrupted files.
3. **dashboard.rs** — extracted a `truncate_left()` helper, now shared by the completion renderer and the recents list (DRY).

Verified: full clean rebuild for `wasm32-wasip1`, zero warnings. Over to you for the merge 🙏

PR #198, opened 2026-06-09
fix(dashboard): embed event_map fallback so stale config can't break Status (#197)

### LINKED ISSUE #197: Dashboard Status column stuck at "-" after update (stale agents-defaults.toml loses event_map)
## Symptom
The dashboard Status column no longer updates — every agent stays at `-` (Unknown), even though hooks fire correctly (confirmed via `/tmp/lince-dashboard/hook-debug.log` showing `PreToolUse`, `Stop`, `SessionStart`, etc.).

## Root cause
Status relies on per-agent `[agents.*.event_map]` in `agents-defaults.toml` to translate native hook events → canonical states. Without it, `StatusMessage::to_agent_status` maps every native event to `Unknown`.

`update.sh` (commit 1e7dd5af) **preserves** an existing on-disk `~/.config/lince-dashboard/agents-defaults.toml` and never overwrites it ("new upstream keys are not auto-merged" by design). Users who first installed **before** LINCE-122 (commit a1c7ec01) added `event_map` therefore keep an event_map-less file across every update → Status column frozen at `-`.

The shipped file in the repo is correct; only stale on-disk copies are affected.

## Fix
`DashboardConfig::event_map_for` now falls back to a binary-embedded copy of the shipped `agents-defaults.toml` (`include_str!`, parsed once via `OnceLock`) when the on-disk map is empty. Built-in agents (claude/codex/pi/opencode) always get their event_map regardless of how stale the on-disk file is; a non-empty on-disk map still overrides.

## Follow-up (separate)
Design inconsistency: docs say `agents-defaults.toml` is "overwritten on updates, customize in config.toml", but update.sh preserves it AND the `lince-add-supported-agent` skill writes custom agents into it. Tracked by backlog lince-117.

maeste:
## Problem
The dashboard Status column stopped updating — every agent stays at `-` (Unknown) even though hooks fire correctly (verified via `/tmp/lince-dashboard/hook-debug.log`).

Native hook events (`PreToolUse`, `Stop`, …) are translated to canonical states via per-agent `[agents.*.event_map]` in `agents-defaults.toml`. `update.sh` (1e7dd5af) **preserves** an existing on-disk `~/.config/lince-dashboard/agents-defaults.toml` and never overwrites it. Installs predating LINCE-122 (a1c7ec01, which introduced `event_map`) therefore keep an event_map-less file across updates → every event maps to `Unknown` → Status frozen at `-`.

## Fix
`DashboardConfig::event_map_for` falls back to a binary-embedded copy of the shipped `agents-defaults.toml` (`include_str!` parsed once via `OnceLock`) when the on-disk map is empty.

Resolution order:
1. on-disk map for the exact agent type (when non-empty) — user overrides still win
2. embedded shipped map for the exact agent type
3. embedded shipped map for the base name (sandbox variants)

Built-in agents (claude/codex/pi/opencode) always resolve their event_map regardless of how stale the on-disk file is.

## Tests
Added: embedded coverage for built-in agents, empty-config fallback (the bug), base-name fallback, and config-overrides-embedded precedence.

## Verification notes
- `cargo check` (host) passes. The WASM build (`cargo build --target wasm32-wasip1`) and in-crate `cargo test` could **not** be run in my environment (no `wasm32-wasip1` std; host target won't link zellij-tile host fns). The fallback logic was additionally validated by replicating it in a standalone host crate against the real `agents-defaults.toml`. **Please rebuild the WASM plugin before merging.**

## Follow-up (out of scope)
Design inconsistency: docs say `agents-defaults.toml` is "overwritten on updates, customize in config.toml", but update.sh preserves it AND the `lince-add-supported-agent` skill writes custom agents into it. Tracked by backlog lince-117.

Closes #197

### DIFF
--- a/lince-dashboard/plugin/src/config.rs
+++ b/lince-dashboard/plugin/src/config.rs
@@ -291,7 +291,12 @@ pub struct DashboardConfig {
     pub agent_layout: AgentLayout,
     #[serde(default)]
     pub focus_mode: FocusMode,
+    // Retained for config back-compat (existing configs may set
+    // `status_method = "file"`), but no longer drives behavior: the dashboard
+    // always polls the .state files now (the pipe is a best-effort fast-path
+    // that can't cross the agent sandbox boundary). See the Timer handler.
     #[serde(default)]
+    #[allow(dead_code)]
     pub status_method: StatusMethod,
     #[serde(default = "default_status_file_dir")]
     pub status_file_dir: String,
--- a/lince-dashboard/plugin/src/config.rs
+++ b/lince-dashboard/plugin/src/config.rs
@@ -1150,12 +1150,57 @@ impl DashboardConfig {
     }
 
     /// Return the event_map for a given agent type, or None if empty/missing.
+    ///
+    /// Resolution order:
+    /// 1. The on-disk config's map for this exact agent type (when non-empty).
+    /// 2. The binary-embedded shipped defaults (exact agent type, then base
+    ///    name).
+    ///
+    /// The embedded fallback exists because `agents-defaults.toml` is shipped
+    /// defaults, but `update.sh` *preserves* a pre-existing on-disk copy (it
+    /// never overwrites it once present). Users who first installed before
+    /// LINCE-122 added `event_map` therefore keep an event_map-less file across
+    /// updates, which left every native hook event mapping to `Unknown` and
+    /// froze the Status column at "-". Falling back to the compiled-in defaults
+    /// keeps status working for the built-in agents no matter how stale the
+    /// on-disk file is, while still letting a non-empty on-disk map override.
     pub fn event_map_for(&self, agent_type: &str) -> Option<&HashMap<String, String>> {
-        self.agent_types
+        if let Some(m) = self
+            .agent_types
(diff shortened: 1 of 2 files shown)


maeste:
## Update: second commit — the actual delivery bug

The embedded event_map fallback (first commit) was necessary but **not sufficient**: status stayed `-` for sandboxed agents because the status event never reached the plugin.

**Root cause of delivery failure:** status was delivered only via `zellij pipe`, but the hook runs *inside* the bwrap agent sandbox and `zellij pipe` can't reach Zellij's IPC socket from there — only `/tmp` is bind-mounted, not the socket dir — so the pipe is silently dropped. The hooks also write `.state` files to the host-shared `/tmp/lince-dashboard`, and the dashboard can read them via `run_command` (works since #143), but that polling was gated behind `status_method == "file"` (default is `pipe`), so the one reliable channel never ran.

**Fix:** poll the `.state` files unconditionally on the timer. The pipe remains an instant fast-path where it works (unsandboxed agents); polling reconciles everyone else within the timer period. Together with the embedded event_map fallback, sandboxed agents now report status.

PR #272, opened 2026-07-13
(feat) installation support for goose

dcolazin:
Added support for goose


maeste:
Thanks for this, @dcolazin — goose is a welcome addition, and the overall shape is right: the registry entry follows the Tier B contract, the level fragments mirror the sibling agents, and the wizard wiring is clean.

We walked the PR through a structured review and will take it from here with a few follow-up commits on the branch, so no action is needed from you. For transparency, what we found and what we'll do:

- **Registry artifact**: `registry.d/goose.toml` is missing the `[sandbox.levels.*]` tables that `gen_registry.py` derives from the new fragments, so `test_registry_sync.py` fails as-committed. We'll rebase onto current `main` (the branch predates the bob/m2 changes — the merge conflict on `SHIPPED_AGENTS` needs both agents kept) and regenerate.
- **goose-paranoid + Anthropic**: goose reads its Anthropic endpoint from `ANTHROPIC_HOST`, not `ANTHROPIC_BASE_URL`, and the credential proxy strips `ANTHROPIC_API_KEY` from the env — so the Anthropic path in paranoid can't currently connect (OpenAI works, since goose honors `OPENAI_BASE_URL`). We'll correct the fragment comment to declare openai-only and open an issue to add `ANTHROPIC_HOST` to the proxy rules.
- **nono profiles**: the nono spawn path reuses the sandboxed launcher as the inner command for agents not hardcoded in `agent.rs`, so nono would nest `agent-sandbox` inside `nono` (this also silently affects bob, which we'll file separately). Since nono is deprecated, we'll drop the three `lince-goose*.json` profiles from this PR rather than ship a path that can't work.
- **Small cleanups we'll apply**: the quickstart `AGENT_SELECTED` array needs a 7th entry after the merge, the `GSM` template label → `GS+` for consistency, and the template blocks get the `env_vars` passthrough plus a note on the three persisted state dirs.

Again — nothing needed from your side; thanks for the clean groundwork.

PR #275, opened 2026-07-25
Ctrl+C Signal Handling

### LINKED ISSUE #274: Ctrl+C during agent "thinking" might terminate the entire sandbox instead of cancelling the current operation
## Summary

When an agent that handles `SIGINT` gracefully (e.g. **goose**) runs inside
`agent-sandbox`, pressing `Ctrl+C` while the agent is "thinking" (processing a
response, executing tools) **kills the entire sandbox** instead of cancelling
the current operation and returning to the agent's prompt.

At the prompt, `Ctrl+C` works correctly (shows "press again to exit"
confirmation). The breakage only occurs when the agent is actively working.

---

## Steps to reproduce

1. Configure goose as an agent in `agent-sandbox` (any sandbox level).
2. Launch the agent:
   ```bash
   agent-sandbox run --agent goose
   ```
3. Type a prompt and press Enter — the agent starts "thinking".
4. Press `Ctrl+C`.

### Expected behavior

The agent cancels the current LLM request / tool execution and returns to the
prompt interface. The sandbox stays alive.

### Actual behavior

The entire sandbox terminates immediately. Output:
```
Sandbox terminated.
```

The user is dropped back to the shell / zellij pane with no agent running.

---

### The process tree

```
zellij PTY (/dev/pts/0)
  └─ agent-sandbox (Python)           ← parent; in the foreground process group
       └─ bwrap (PID 1 in namespace)   ← --unshare-pid --die-with-parent
            └─ goose session (PID 2)   ← the agent; catches SIGINT
```

All three processes share the **same process group** (no `start_new_session`,
no `--new-session` flag on bwrap).

### Why the prompt works but thinking doesn't

| Phase | Terminal mode | What `Ctrl+C` does |
|---|---|---|
| **At the prompt** | Goose sets **raw mode** (`ISIG=0`) | `Ctrl+C` → byte `0x03` → goose's TUI catches it as a key event. **No SIGINT is generated.** agent-sandbox never sees it. |
| **During thinking** | Goose restores **cooked mode** (`ISIG=1`) | `Ctrl+C` → kernel generates **SIGINT** → delivered to **all** processes in the foreground process group. |

Confirmed at runtime by reading the live termios attributes of goose's stdin
(`/dev/pts/0`): `ICANON=1`, `ISIG=1`, `ECHO=1`.

### The fatal race

When `SIGINT` is generated during thinking, all three processes receive it
simultaneously:

| Process | What happens | Speed |
|---|---|---|
| `agent-sandbox` (Python) | `KeyboardInterrupt` raised → `except KeyboardInterrupt: print("Sandbox terminated.")` → **exits** | **Instant** (C-level signal handler) |
| `bwrap` (PID 1) | SIGINT **ignored** — PID 1 in a PID namespace has `SIGNAL_UNKILLABLE`; bwrap has no custom handler (`SigCgt: 0x0`) | N/A |
| `goose` (PID 2) | Custom SIGINT handler fires (`SigCgt: 0x100000442`) → tokio signal task wakes → cancels LLM request | **Slower** (async runtime cooperation) |

agent-sandbox exits first. When it dies:

1. **bwrap `--die-with-parent`** detects that its parent (agent-sandbox) is gone.
2. bwrap kills all children in the sandbox → **goose is killed** mid-interrupt-handling.
3. The entire sandbox is destroyed.

Goose never gets to finish handling the `SIGINT`.

### The code at fault

All three backend paths in `sandbox/agent-sandbox` have the same pattern:

```python
# bwrap backend (~line 5152), nono backend (~line 4339), seatbelt backend (~line 4631)
try:
    result = subprocess.run(cmd, cwd=str(project_dir))
    sys.exit(result.returncode)
except KeyboardInterrupt:
    print("\nSandbox terminated.")   # ← agent-sandbox exits here, killing everything
```

The `learn` command (~line 7345) is even worse: it explicitly sends `SIGTERM`
to the child on `KeyboardInterrupt`.

---

### Fix 1 (recommended): `Popen` + `_sandbox_wait` with `graceful_interrupt` config flag

Replace `subprocess.run()` + `except KeyboardInterrupt` with a `Popen` +
`_sandbox_wait()` helper that respects a new **`graceful_interrupt`** config
option (in `[security]`, default `true`).

**When `graceful_interrupt = true` (default):**

| User action | agent-sandbox | Child (goose) |
|---|---|---|
| First `Ctrl+C` | Catches `KeyboardInterrupt`, **does not exit**, loops back to `proc.wait()` | Receives SIGINT via process group → cancels LLM request → returns to prompt |
| Second `Ctrl+C` within 3 s | Force-terminates: `SIGTERM` → 5 s timeout → `SIGKILL` | Killed |
| Second `Ctrl+C` after 3 s | Reset to "first interrupt" — graceful again | Handles SIGINT again |
| `/exit` or double-`Ctrl+C` at prompt | `proc.wait()` returns normally → clean exit | Exits normally |

**When `graceful_interrupt = false` (legacy):**

| User action | agent-sandbox | Child |
|---|---|---|
| `Ctrl+C` | Prints "Sandbox terminated." → exits immediately | Killed via `--die-with-parent` |

The 3-second reset window ensures two `Ctrl+C` presses far apart (e.g. one
during a thinking phase, then another minutes later during a new thinking
phase) are treated as independent graceful interrupts, not as a force-quit.

This design provides two escape hatches for misbehaving agents:
1. **Double `Ctrl+C`** within 3 seconds forces termination (SIGTERM → SIGKILL).
2. **Config flag** `graceful_interrupt = false` reverts to legacy immediate-exit.

Helper function:

```python
_SANDBOX_INTERRUPTED = object()   # sentinel for legacy-mode interrupt

def _sandbox_wait(proc, graceful_interrupt):
    """Wait for a sandboxed child process, handling Ctrl+C."""

    if not graceful_interrupt:
        # Legacy: first Ctrl+C terminates immediately
        try:
            return proc.wait()
        except KeyboardInterrupt:
            print("\nSandbox terminated.")
            return _SANDBOX_INTERRUPTED

    _RESET_SECS = 3
    interrupt_count = 0
    last_interrupt = 0.0

    while True:
        try:
            return proc.wait()
        except KeyboardInterrupt:
            now = time.monotonic()
            if now - last_interrupt > _RESET_SECS:
                interrupt_count = 0
            interrupt_count += 1
            last_interrupt = now

            if interrupt_count >= 2:
                # Escape hatch: force-kill misbehaving agents
                print("\nForce terminating sandbox...")
                try:
                    proc.send_signal(signal.SIGTERM)
                    return proc.wait(timeout=5)
                except subprocess.TimeoutExpired:
                    proc.kill()
                    return proc.wait()
            # Graceful: first interrupt — child handles SIGINT via
            # the shared process group. Continue waiting.
            continue
```

Each backend path changes from:

```python
# BEFORE (all backends):
try:
    result = subprocess.run(cmd, cwd=str(project_dir))
    sys.exit(result.returncode)
except KeyboardInterrupt:
    print("\nSandbox terminated.")
```

to:

```python
# AFTER (all backends):
graceful_interrupt = sec.get("graceful_interrupt", True)
try:
    proc = subprocess.Popen(cmd, cwd=str(project_dir))
    result = _sandbox_wait(proc, graceful_interrupt)
    if result is _SANDBOX_INTERRUPTED:
        sys.exit(130)
    sys.exit(result)
except FileNotFoundError as exc:
    ...
```

**Why it works:**

1. `Ctrl+C` → SIGINT → all processes receive it (unchanged).
2. agent-sandbox catches `KeyboardInterrupt` → **does not exit** → loops back to `proc.wait()`.
3. goose catches SIGINT → cancels the LLM request → returns to prompt (goose keeps running).
4. `proc.wait()` continues blocking → agent-sandbox stays alive → no `--die-with-parent` trigger.
5. When goose eventually exits (user types `/exit`, double-presses `Ctrl+C` at the prompt) → `proc.wait()` returns → agent-sandbox exits cleanly.

**Edge cases:**

| Scenario | Behavior |
|---|---|
| Agent handles SIGINT (goose) | Agent cancels current op, returns to prompt. agent-sandbox keeps waiting. ✓ |
| Agent does NOT handle SIGINT | Agent killed by default SIGINT action. `proc.wait()` returns. agent-sandbox exits with child's exit code. ✓ |
| User presses `Ctrl+C` twice quickly | Second press forces termination (SIGTERM → SIGKILL). ✓ |
| User presses `Ctrl+C` twice > 3 s apart | Reset to "first interrupt" — both are graceful. ✓ |
| Misbehaving agent ignores SIGINT | User presses `Ctrl+C` twice → force kill. Or set `graceful_interrupt = false`. ✓ |
| Agent is at the prompt (raw mode) | No SIGINT generated. agent-sandbox unaffected. Agent handles byte `0x03`. ✓ |

**Files to change:**

| Backend | Line (approx.) | Change |
|---|---|---|
| bwrap | ~5152 | `subprocess.run` → `Popen` + `_sandbox_wait` |
| nono | ~4339 | Same |
| seatbelt | ~4631 | Same |
| learn | ~7345 | `proc.wait()` → `_sandbox_wait` (non-duration path) |
| config | — | Add `graceful_interrupt = true` to `[security]` in `_DEFAULT_CONFIG` and `config.toml.example` |
| banner | — | Add `interrupt` line showing graceful/immediate status |

**Config option:**

```toml
[security]
# Graceful Ctrl+C: when true (default), the first Ctrl+C while the agent
# is working does NOT terminate the sandbox. The agent receives SIGINT
# via the shared process group and is expected to handle it (e.g. cancel
# the current LLM request and return to the prompt). A second Ctrl+C
# within 3 seconds forces termination (SIGTERM → SIGKILL).
# Set to false to restore the old behaviour (immediate exit on Ctrl+C).
graceful_interrupt = true
```

**Pros:** Minimal change, correct for all agents, built-in escape hatch, feature-flagged.
**Cons:** None identified.

---

### Fix 2 (alternative): `start_new_session=True` + manual signal forwarding

Put the child in its own session and forward signals explicitly:

```python
proc = subprocess.Popen(cmd, cwd=str(project_dir), start_new_session=True)
signal.signal(signal.SIGINT, lambda *_: os.killpg(proc.pid, signal.SIGINT))
return_code = proc.wait()
```

**Pros:** agent-sandbox becomes the sole signal router — no race conditions.
**Cons:** More invasive; changes process-group semantics for all backends; if agent-sandbox crashes, the child never receives signals.

---

## Environment

| Component | Version / value |
|---|---|
| agent-sandbox | lince `sandbox/agent-sandbox` (installed at `~/.local/bin/agent-sandbox`) |
| goose | v1.43.0 (`~/.local/bin/goose`) |
| Backend | bwrap (Linux, `--unshare-pid --die-with-parent`) |
| Config | `new_session = false` (default) |
| Terminal | `/dev/pts/0`, cooked mode during thinking (`ISIG=1`) |

## Labels

`bug`, `sandbox`, `signal-handling`

dcolazin:
Fixes #274 

## Summary

When an agent that handles `SIGINT` gracefully (e.g. goose) runs inside
`agent-sandbox`, pressing `Ctrl+C` while the agent is working **killed the
entire sandbox** instead of cancelling the current operation and returning
to the prompt.

This PR implements a three-layer fix: `start_new_session=True` to shield
bwrap from terminal SIGINT, `/proc`-based SIGINT forwarding to deliver the
signal to the agent only, and a `_sandbox_wait()` loop with a double-Ctrl+C
force-kill escape hatch.

---

## Problem

```
$ agent-sandbox run --agent goose
> Write me a sorting algorithm    ← type prompt, press Enter
                                  ← agent starts thinking...
^C                                ← Ctrl+C
                                  ← EXPECTED: cancel request, return to prompt
                                  ← ACTUAL: entire sandbox terminates
```

At the prompt, `Ctrl+C` worked correctly (showed "press again to exit").
The breakage only occurred during thinking.

---

### 1. Terminal mode differs between prompt and thinking

| Phase | Terminal mode | `Ctrl+C` behaviour |
|---|---|---|
| At the prompt | Goose sets **raw mode** (`ISIG=0`) | Byte `0x03` → goose TUI handles it. No SIGINT generated. |
| During thinking | Goose restores **cooked mode** (`ISIG=1`) | Kernel generates **SIGINT** → all processes in foreground group receive it. |

### 2. The bwrap process tracked by `Popen` has no signal protection

```
agent-sandbox (Python)      ← foreground process group
  └─ bwrap (original)       ← killed by SIGINT (no SIGNAL_UNKILLABLE)
       └─ bwrap-clone       ← PID 1 in --unshare-pid namespace
            └─ goose        ← catches SIGINT (too late)
```

The original code used `subprocess.run(cmd)` with all processes sharing the
same process group. When SIGINT arrived:

- **bwrap (original)** — killed instantly. No custom handler, no
  `SIGNAL_UNKILLABLE`. Return code `-2`.
- **bwrap-clone** — ignored SIGINT (`SIGNAL_UNKILLABLE` as namespace PID 1).
- **goose** — handler fired, but the sandbox was already being torn down.

When bwrap (original) died → `--die-with-parent` triggered → bwrap-clone
killed goose → sandbox destroyed.

**Key insight:** `SIGNAL_UNKILLABLE` protects bwrap's **cloned child** inside
the PID namespace, NOT the **original bwrap process** that `Popen` tracks.

Debug logging confirmed: `proc.wait() returned -2` (killed by signal 2).

---

### Layer 1 — `start_new_session=True` (shields bwrap)

```python
proc = subprocess.Popen(cmd, cwd=str(project_dir),
                        start_new_session=True)
```

Calls `setsid()` before exec'ing bwrap. bwrap is now in its own session.
Terminal `Ctrl+C` generates SIGINT only for agent-sandbox's foreground group.
**bwrap never receives terminal SIGINT.**

### Layer 2 — `_forward_sigint_to_descendants()` (delivers SIGINT to goose)

With bwrap in a separate session, goose also doesn't receive SIGINT. The
helper traverses `/proc` to find goose's PID and sends SIGINT **only to
descendants of bwrap** (not bwrap itself):

```
bwrap (root_pid)         ← NOT signaled → survives
  └─ bwrap-clone         ← signaled, but SIGNAL_UNKILLABLE → ignores
       └─ goose          ← signaled → catches SIGINT → cancels request
```

### Layer 3 — `_sandbox_wait()` (graceful loop + escape hatch)

```python
while True:
    try:
        return proc.wait()
    except KeyboardInterrupt:
        # First Ctrl+C: forward SIGINT to goose, keep waiting
        _forward_sigint_to_descendants(proc.pid)
        continue
        # Second Ctrl+C within 3s: SIGTERM → SIGKILL
```

### Feature flag

```toml
[security]
graceful_interrupt = true   # default; set false for legacy immediate-exit
```

---

### During thinking — fixed

```
Ctrl+C → SIGINT → agent-sandbox only (bwrap in separate session)

  ① _sandbox_wait catches KeyboardInterrupt (interrupt_count = 1)
  ② _forward_sigint_to_descendants(bwrap.pid)
       → bwrap-clone: ignored (SIGNAL_UNKILLABLE)
       → goose: SIGINT caught → cancels LLM request → returns to prompt
  ③ continue waiting → sandbox stays alive ✓
```

### At the prompt — unchanged

```
Raw mode (ISIG=0) → Ctrl+C is byte 0x03 → goose TUI handles it
No SIGINT → agent-sandbox unaffected ✓
```

### Double Ctrl+C — force-kill

```
Second Ctrl+C within 3s → SIGTERM → bwrap → --die-with-parent → goose killed ✓
```

---

## Edge Cases

| Scenario | Behaviour |
|---|---|
| Agent handles SIGINT (goose) | Cancels request, returns to prompt, sandbox alive ✓ |
| Agent does NOT handle SIGINT | Default SIGINT action kills agent, `proc.wait()` returns, clean exit ✓ |
| Agent ignores SIGINT (misbehaving) | First Ctrl+C: no visible effect. Second: force-kill. Or `graceful_interrupt = false` ✓ |
| `/proc` unavailable (non-Linux) | Forward fails silently, sandbox alive, double Ctrl+C to force-kill ✓ |
| Process already exited when forwarding | `os.kill` → `ProcessLookupError` → caught, `proc.wait()` returns ✓ |
| Sandbox crashes / parent killed | Agent terminated via `--die-with-parent` + PID namespace teardown ✓ |
| Two Ctrl+C within 3s | Force-kill (SIGTERM → SIGKILL) ✓ |
| Two Ctrl+C > 3s apart | Counter resets, both treated as graceful ✓ |
| bwrap-clone receives forwarded SIGINT | Ignored — `SIGNAL_UNKILLABLE` as namespace PID 1 ✓ |
| Goose terminal access in separate session | fd 0 inherited; raw mode passes bytes to any reader; no SIGTTIN (different session) ✓ |

---

## Files Changed

| File | Change |
|---|---|
| `sandbox/agent-sandbox` | `_forward_sigint_to_descendants()` + `_sandbox_wait()` + `start_new_session=True` on all 3 backends (bwrap, nono, seatbelt) + `graceful_interrupt` config read + banner output + `cmd_learn` update |
| `sandbox/config.toml.example` | New `graceful_interrupt` option in `[security]` |

### Diff summary

```
 sandbox/agent-sandbox       | ~130 insertions, ~14 deletions
 sandbox/config.toml.example | +8 lines
```

No dependencies added. No existing API changed. The `graceful_interrupt`
flag defaults to `true` but can be set to `false` for instant rollback.

---

## Testing

Tested manually with goose v1.43.0 on Linux (bwrap backend):

| Test | Result |
|---|---|
| `Ctrl+C` during thinking | ✅ Request cancelled, goose returns to prompt, sandbox alive |
| `Ctrl+C` at the prompt | ✅ Shows "press again to exit" (unchanged) |
| `Ctrl+C` twice within 3s | ✅ Force-terminates sandbox (SIGTERM → SIGKILL) |
| `Ctrl+C` twice > 3s apart | ✅ Both treated as graceful (no accidental force-kill) |
| Normal `/exit` | ✅ Clean shutdown (unchanged) |
| `graceful_interrupt = false` | ✅ Legacy behaviour — immediate exit on first Ctrl+C |

### DIFF
--- a/sandbox/agent-sandbox
+++ b/sandbox/agent-sandbox
@@ -205,6 +205,14 @@ unshare_pid = true
 # May interfere with Ctrl+C propagation. Disable if you have issues.
 new_session = false
 
+# Graceful Ctrl+C: when true (default), the first Ctrl+C while the agent
+# is working does NOT terminate the sandbox. The agent receives SIGINT
+# via the shared process group and is expected to handle it (e.g. cancel
+# the current LLM request and return to the prompt). A second Ctrl+C
+# within 3 seconds forces termination (SIGTERM → SIGKILL).
+# Set to false to restore the old behaviour (immediate exit on Ctrl+C).
+graceful_interrupt = true
+
 # Block git push via a wrapper script
 block_git_push = true
 
@@ -4158,6 +4166,121 @@ def landlock_plan(config: dict, sandbox_level: str | None,
     }
 
 
+# Sentinel returned by _sandbox_wait when the sandbox is terminated by a
+# user interrupt in legacy (non-graceful) mode.
+_SANDBOX_INTERRUPTED = object()
+
+
+def _forward_sigint_to_descendants(root_pid: int) -> bool:
+    """Send SIGINT to all descendant processes of *root_pid*.
+
+    With ``start_new_session=True`` the bwrap process (root_pid) is in its
+    own session and does not receive terminal SIGINT.  We need to forward
+    the signal to the agent (e.g. goose) so it can cancel the current
+    operation.
+
+    The process tree is::
+
+        bwrap (root_pid) → bwrap-clone (PID 1 in namespace) → goose
+
+    bwrap-clone has ``SIGNAL_UNKILLABLE`` and ignores SIGINT.
+    goose catches SIGINT and handles it (cancel request, return to prompt).
+    root_pid (bwrap) itself is **not** signaled, so it survives.
+
+    Returns True if at least one descendant was signaled.
+    """
+    import glob
+
+    # Build parent → children map from /proc
+    children_of: dict[int, list[int]] = {}
+    for stat_path in glob.glob("/proc/[0-9]*/stat"):
+        try:
+            with open(stat_path) as fh:
+                data = fh.read()
+            # Format: "pid (comm) state ppid ..."
+            # comm can contain spaces and parens.
+            lparen = data.find("(")
+            rparen = data.rfind(")")
+            pid = int(data[:lparen].strip())
+            fields = data[rparen + 2:].split()
+            ppid = int(fields[1])
+            children_of.setdefault(ppid, []).append(pid)
+        except (OSError, ValueError, IndexError):
+            continue
+
+    # BFS to collect all descendant PIDs (not root_pid itself)
+    to_signal: list[int] = []
+    queue = list(children_of.get(root_pid, []))
+    while queue:
+        pid = queue.pop(0)
+        to_signal.append(pid)
+        queue.extend(children_of.get(pid, []))
+
+    if not to_signal:
+        return False
+
+    for pid in to_signal:
+        try:
+            os.kill(pid, signal.SIGINT)
+        except OSError:
+            pass
+    return True
+
+
+def _sandbox_wait(proc: subprocess.Popen, graceful_interrupt: bool):
+    """Wait for a sandboxed child process, handling Ctrl+C.
+
+    When *graceful_interrupt* is True, the first Ctrl+C does **not**
+    terminate the sandbox.  The child already received ``SIGINT`` via the
+    shared process group and is expected to handle it (e.g. cancel an LLM
+    request and return to the prompt).  A second Ctrl+C within three
+    seconds forces termination (``SIGTERM`` → ``SIGKILL``).
+
+    When *graceful_interrupt* is False, Ctrl+C terminates the sandbox
+    immediately (legacy behaviour).
+
+    Returns the child's exit code on normal or forced termination.
+    Returns ``_SANDBOX_INTERRUPTED`` when the sandbox is terminated by a
+    user interrupt in legacy mode.
+    """
+    if not graceful_interrupt:
+        try:
+            return proc.wait()
+        except KeyboardInterrupt:
+            print("\nSandbox terminated.")
+            return _SANDBOX_INTERRUPTED
+
+    _RESET_SECS = 3
+    interrupt_count = 0
+    last_interrupt = 0.0
+
+    while True:
+        try:
+            return proc.wait()
+        except KeyboardInterrupt:
+            now = time.monotonic()
+            if now - last_interrupt > _RESET_SECS:
+                interrupt_count = 0
+            interrupt_count += 1
+            last_interrupt = now
+
+            if interrupt_count >= 2:
+                print("\nForce terminating sandbox...")
+                try:
+                    proc.send_signal(signal.SIGTERM)
+                    return proc.wait(timeout=5)
+                except subprocess.TimeoutExpired:
+                    proc.kill()
+                    return proc.wait()
+            # First Ctrl+C: the child is in its own session (start_new_session)
+            # so it did NOT receive terminal SIGINT.  Forward SIGINT to the
+            # agent (goose) by signaling bwrap's descendants — bwrap-clone
+            # (PID 1 in namespace) ignores it via SIGNAL_UNKILLABLE, while
+            # goose catches it and can cancel the current request.
+            _forward_sigint_to_descendants(proc.pid)
+            continue
+
+
 def cmd_run(args, agent_extra: list[str]) -> None:
     config, config_path = load_config()
 
@@ -4238,6 +4361,7 @@ def cmd_run(args, agent_extra: list[str]) -> None:
     profile = provider
 
     sec = config.get("security", {})
+    graceful_interrupt = sec.get("graceful_interrupt", True)
 
     # --- nono backend path (simpler — nono handles isolation itself) -------
     if backend == "nono":
@@ -4301,6 +4425,7 @@ def cmd_run(args, agent_extra: list[str]) -> None:
         print(f"  isolation   Landlock LSM / Seatbelt")
         print(f"  network     enabled")
         print(f"  git push    {'blocked' if sec.get('block_git_push', True) else 'allowed'}")
+        print(f"  interrupt   {'graceful (2×Ctrl+C to force)' if graceful_interrupt else 'immediate exit'}")
         print(
             f"  permissions "
             f"{'ask (safe mode)' if args.safe else 'per agent defaults'}"
@@ -4336,10 +4461,12 @@ def cmd_run(args, agent_extra: list[str]) -> None:
             )
 
         try:
-            result = subprocess.run(cmd, cwd=str(project_dir))
-            sys.exit(result.returncode)
-        except KeyboardInterrupt:
-            print("\nSandbox terminated.")
+            proc = subprocess.Popen(cmd, cwd=str(project_dir),
+                                    start_new_session=True)
+            result = _sandbox_wait(proc, graceful_interrupt)
+            if result is _SANDBOX_INTERRUPTED:
+                sys.exit(130)
+            sys.exit(result)
         except FileNotFoundError as exc:
             print(f"Error: {exc}", file=sys.stderr)
             sys.exit(1)
@@ -4580,6 +4707,7 @@ def cmd_run(args, agent_extra: list[str]) -> None:
         if sb_proxy:
             print(f"  cred proxy  127.0.0.1:{sb_proxy_port} (direct TCP)")
         print(f"  git push    {'blocked' if sec.get('block_git_push', True) else 'allowed'}")
+        print(f"  interrupt   {'graceful (2×Ctrl+C to force)' if graceful_interrupt else 'immediate exit'}")
         print(
             f"  permissions "
             f"{'ask (safe mode)' if args.safe else 'per agent defaults'}"
@@ -4628,10 +4756,12 @@ def cmd_run(args, agent_extra: list[str]) -> None:
         write_policy_record(args.id, sb_record)
 
         try:
-            result = subprocess.run(cmd, cwd=str(project_dir))
-            sys.exit(result.returncode)
-        except KeyboardInterrupt:
-            print("\nSandbox terminated.")
+            proc = subprocess.Popen(cmd, cwd=str(project_dir),
+                                    start_new_session=True)
+            result = _sandbox_wait(proc, graceful_interrupt)
+            if result is _SANDBOX_INTERRUPTED:
+                sys.exit(130)
+            sys.exit(result)
         except FileNotFoundError as exc:
             print(f"Error: {exc}", file=sys.stderr)
             sys.exit(1)
@@ -5050,6 +5180,7 @@ def cmd_run(args, agent_extra: list[str]) -> None:
         print(f"  landlock    fs only (ABI {landlock_abi})")
     print(f"  pid isolat. {'on' if sec.get('unshare_pid', True) else 'off'}")
     print(f"  git push    {'blocked' if sec.get('block_git_push', True) else 'allowed'}")
+    print(f"  interrupt   {'graceful (2×Ctrl+C to force)' if graceful_interrupt else 'immediate exit'}")
     print(f"  env vars    clearenv + whitelist")
     if proxy:
         targets = [
@@ -5149,16 +5280,18 @@ def cmd_run(args, agent_extra: list[str]) -> None:
     write_policy_record(args.id, bwrap_record)
 
     try:
-        result = subprocess.run(cmd, cwd=str(project_dir))
+        proc = subprocess.Popen(cmd, cwd=str(project_dir),
+                                start_new_session=True)
+        result = _sandbox_wait(proc, graceful_interrupt)
+        if result is _SANDBOX_INTERRUPTED:
+            sys.exit(130)
         snap_conf = _get_snapshot_config(config)
         if snap_conf["auto_config"] or snap_conf["auto_project"]:
             print(
                 "\nTip: run 'agent-sandbox snapshot-diff' to see "
                 "what changed since the pre-run snapshot.",
             )
-        sys.exit(result.returncode)
-    except KeyboardInterrupt:
-        print("\nSandbox terminated.")
+        sys.exit(result)
     except FileNotFoundError as exc:
         print(f"Error: {exc}", file=sys.stderr)
         sys.exit(1)
@@ -7247,6 +7380,7 @@ def cmd_learn(args) -> None:
     # 1. Load config and check backend first
     config, config_path = load_config()
     project_dir = Path.cwd().resolve()
+    graceful_interrupt = config.get("security", {}).get("graceful_interrupt", True)
 
     # 1a. Gate: learn mode requires strace (Linux-only). Seatbelt and nono
     # backends cannot use strace-based learn mode.
@@ -7359,10 +7493,10 @@ def cmd_learn(args) -> None:
                     proc.kill()
                     proc.wait()
         else:
-            proc.wait()
+            _sandbox_wait(proc, graceful_interrupt)
+            proc.poll()  # ensure returncode is set
 
     except KeyboardInterrupt:
-        print("\n\nInterrupted, terminating agent...")
         if proc and proc.poll() is None:
             proc.send_signal(signal.SIGTERM)
             try:
--- a/sandbox/config.toml.example
+++ b/sandbox/config.toml.example
@@ -145,6 +145,14 @@ unshare_pid = true
 # May interfere with Ctrl+C propagation. Disable if you have issues.
 new_session = false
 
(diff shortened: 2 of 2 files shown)


sandbox/agent-sandbox
```diff
@@ -4158,6 +4166,121 @@ def landlock_plan(config: dict, sandbox_level: str | None,
...
+    seconds forces termination (``SIGTERM`` → ``SIGKILL``).
+
+    When *graceful_interrupt* is False, Ctrl+C terminates the sandbox
+    immediately (legacy behaviour).
+
+    Returns the child's exit code on normal or forced termination.
+    Returns ``_SANDBOX_INTERRUPTED`` when the sandbox is terminated by a
+    user interrupt in legacy mode.
+    """
+    if not graceful_interrupt:
```
maeste, line 4246: With `start_new_session=True` applied unconditionally at all three spawn sites, this legacy branch no longer restores the old behavior: on Ctrl+C the sandbox tree is in its own session and receives no signal at all, and this branch just prints and returns — it never kills the child.

- On bwrap, `--die-with-parent` masks this (the sandbox is killed when agent-sandbox exits).
- On seatbelt and nono there is no parent-death mechanism, so the agent survives agent-sandbox's exit as an orphan still attached to the terminal. The old `subprocess.run` killed the child on `KeyboardInterrupt`; this branch replaces that kill with nothing.

Suggested fix: on `KeyboardInterrupt` in this branch, `proc.terminate()`/`proc.kill()` + `proc.wait()` before returning — i.e. restore the old signal delivery and reaping, not just the print. As written, the flag documented as "restore the old behaviour (immediate exit on Ctrl+C)" cannot deliver that contract on two of the three backends.

sandbox/agent-sandbox
```diff
@@ -4158,6 +4166,121 @@ def landlock_plan(config: dict, sandbox_level: str | None,
...
+                    return proc.wait(timeout=5)
+                except subprocess.TimeoutExpired:
+                    proc.kill()
+                    return proc.wait()
+            # First Ctrl+C: the child is in its own session (start_new_session)
+            # so it did NOT receive terminal SIGINT.  Forward SIGINT to the
+            # agent (goose) by signaling bwrap's descendants — bwrap-clone
+            # (PID 1 in namespace) ignores it via SIGNAL_UNKILLABLE, while
+            # goose catches it and can cancel the current request.
+            _forward_sigint_to_descendants(proc.pid)
```
maeste, line 4280: The return value is discarded, and when the forward signals nothing the first Ctrl+C is a completely silent no-op — which is always the case on macOS (no `/proc`: `glob` returns `[]`, this returns `False`, no output anywhere).

Combined with the 3s counter reset above, a user who presses Ctrl+C, sees nothing, waits a few seconds and presses again (the natural pattern) can never force-terminate: every press spaced more than 3s apart resets the counter to 1. Pre-PR, one Ctrl+C terminated the sandbox; this is now the default-on path on the seatbelt (macOS) backend, while the banner promises "graceful".

Suggested fix: when the forwarder returns `False` (or unconditionally), print a short hint (e.g. "nothing to interrupt — press Ctrl+C again to force exit"); consider also gating graceful mode per-backend, since it can only work where `/proc` exists.

maeste (changes requested):
Thanks for the careful root-cause analysis in #274 and for documenting the pivot from the issue's Fix 1 to the session-isolation hybrid — the `/proc` forwarding design is sound and the bwrap path works as described. Requesting changes for two majors (both about the flag's contract, not the core bwrap fix — see inline comments):

1. `graceful_interrupt = false` doesn't restore the old behavior and orphans the agent on seatbelt/nono (no parent-death mechanism there; the legacy branch never kills the child).
2. The first Ctrl+C is a silent no-op whenever forwarding signals nothing — always the case on macOS — and presses spaced >3s apart can never force-terminate.

Also confirmed in review, non-blocking:
- a third Ctrl+C during the 5s SIGTERM grace escapes `_sandbox_wait` with a traceback and skips the SIGKILL fallback;
- `cmd_learn` discards the `_sandbox_wait` result (legacy-mode trace parsing races strace's teardown) and its `Popen` has no `start_new_session`, so graceful mode can't trigger there;
- the comment in `config.toml.example` and the `_sandbox_wait` docstring still describe the shared-process-group mechanism this PR removes — worth reconciling together with the schema/config-contract chain (`lince-config` schema → generated docs), since two local contract tests currently fail on the new key.

Happy to re-review once the two majors are addressed.

maeste:
@dcolazin any chance to have you back on this great contribution after my review. I know I disappeared for long and you may be pissed of on waiting for the review :)

dcolazin:
I'm sorry, this September is very busy :) I had produced some code, but didn't had the time to test it. I have now pushed it on my branch.

maeste:
Thanks again for this contribution @dcolazin — the three-layer design and the way you addressed every finding from my review round are excellent work. 🙇

While re-reviewing I found four further issues (two major) in the interaction between `start_new_session=True` and the rest of the sandbox, so to unblock the PR I pushed two maintainer-side commits on top of your branch. Your commits are untouched — the design and the implementation credit stays entirely with you:

1. **paranoid egress bridge killed by the first graceful Ctrl+C** (major) — the socat bridge is forked by the same PID that then execs bwrap, so the "signal all descendants" forward killed it: the agent survived but every later API call hit ECONNREFUSED. The forward now skips host-side bridge helpers (`_BRIDGE_COMMS`).
2. **closing the terminal/pane orphaned the sandbox** (major) — with the sandbox in its own session, SIGHUP killed agent-sandbox while the detached tree kept running unsupervised (seatbelt/nono have no `--die-with-parent`). A SIGHUP handler now tears the tree down, restores the terminal, and re-raises with the conventional exit status.
3. **SIGWINCH no longer delivered** — resizes are now forwarded into the sandbox so full-screen agent TUIs redraw correctly.
4. **legacy mode wasn't "immediate exit"** — docs now describe the actual SIGINT → SIGTERM → SIGKILL escalation.

Also in the follow-ups: signal-death exit codes map to the conventional 128+n (143/137 instead of 241/247), and the terminal state captured before the spawn is restored on every kill path (an agent dying in raw mode no longer leaves the terminal broken). The second commit merges current `main` to keep the PR mergeable.

Validated with the repo suites plus functional checks against real socat/bwrap trees, including an end-to-end graceful Ctrl+C through real bwrap with a signal-catching agent: the agent cancels, exits 0, and the sandbox winds down cleanly.

maeste (approved):
Approving: the three-layer design is sound, every review finding is addressed, and the maintainer-side follow-ups are verified end to end. All credit to @dcolazin for the original implementation and the review fixes.

dcolazin:
Thank you for the review and the additional fixes!

PR #276, opened 2026-07-31
test(sandbox): pin sys.platform for landlock_plan tests (#242)

### LINKED ISSUE #242: test_policy_record: make landlock_plan tests portable on non-Linux
Follow-up from #211.

I hit this while checking the merged effective-policy / Landlock path locally on macOS/non-Linux.

Command:

```bash
python3 scripts/tests/test_policy_record.py
```

Observed:

```text
Ran 13 tests in 0.002s

FAILED (errors=3)
```

The failing tests are:

- `test_landlock_plan_keeps_bind_ports_without_proxy`
- `test_landlock_plan_normal_open_has_no_net_rules`
- `test_landlock_plan_paranoid_denies_bind_allows_bridge`

All three fail with the same shape:

```text
TypeError: 'NoneType' object is not subscriptable
```

The common cause appears to be that these tests call `landlock_plan(...)` directly and then index into the returned plan, while `sandbox/agent-sandbox` correctly returns `None` from `landlock_plan(...)` when `sys.platform != "linux"`.

That runtime behavior looks right for non-Linux. The issue is test portability: these plan-construction tests assume a Linux platform even though the rest of `scripts/tests/test_policy_record.py` is useful to run from macOS or another non-Linux dev/CI host.

Expected shape:

- record-shape tests remain runnable on non-Linux; and
- Linux-specific Landlock plan semantics are either tested under a mocked Linux platform or marked Linux-only.

Possible fixes:

- patch/mock `sys.platform` to `"linux"` around the `landlock_plan(...)` plan-construction tests so they test plan semantics without depending on the host OS; or
- split those three tests into a Linux-only class and leave the rest of `test_policy_record.py` cross-platform.

Not a blocker for the Linux Landlock implementation. This is just a way to keep non-Linux/local CI from reporting false failures while checking the effective-policy record work.

oscerd:
Closes #242.

## What changed

`landlock_plan()` returns `None` when `sys.platform != "linux"` — correct runtime behaviour — but the three plan-construction tests in `scripts/tests/test_policy_record.py` index straight into the result, so they raise `TypeError: 'NoneType' object is not subscriptable` on a macOS dev host while the rest of the file is perfectly cross-platform.

This pins `sys.platform` to `"linux"` on exactly those three tests via `mock.patch`.

I chose mocking over marking them Linux-only (the issue offers both) because the plan is pure dict arithmetic with no syscalls — so mocking exercises the real semantics from any host, whereas a skip would mean macOS/non-Linux CI never covers that logic at all.

**The scoping is deliberate, not incidental.** `probe_landlock_abi()` gates a real `ctypes` syscall (`__NR_landlock_create_ruleset` = 444) on the same `sys.platform` flag, and `build_policy_record()` calls it. A class-wide patch would therefore make a non-Linux host dial syscall 444 on Darwin, where that number means something else entirely. The patch stays on the three plan tests, which never touch the probe.

Also adds `test_landlock_plan_is_none_off_linux`, so the guard the other three patch away is itself covered — otherwise this change would silently remove the only assertion that no Landlock shim gets planned off-Linux.

## How to test

On Linux:

```bash
python3 scripts/tests/test_policy_record.py   # Ran 14 tests — OK
```

To verify the non-Linux path without a Mac, pin the platform before the module loads:

```python
# as_darwin.py
import runpy, sys
sys.platform = "darwin"
sys.argv = [sys.argv[1], *sys.argv[2:]]
runpy.run_path(sys.argv[0], run_name="__main__")
```

```bash
python3 as_darwin.py scripts/tests/test_policy_record.py
```

| | before | after |
|---|---|---|
| Linux | Ran 13 — OK | Ran 14 — OK |
| simulated macOS | Ran 13 — **FAILED (errors=3)** | Ran 14 — OK |

The "before" column reproduces the issue report exactly.

## Backward compatibility

Test-only change; no runtime code touched. `ruff check` on the file reports the same 4 pre-existing findings before and after (EXE001, I001, C408, SIM117) — left alone to keep the PR focused.

### DIFF
--- a/scripts/tests/test_policy_record.py
+++ b/scripts/tests/test_policy_record.py
@@ -14,6 +14,7 @@ import os
 import tempfile
 import unittest
 from pathlib import Path
+from unittest import mock
 
 REPO_ROOT = Path(__file__).resolve().parent.parent.parent
 
@@ -101,6 +102,14 @@ class TestRecordHonesty(unittest.TestCase):
     """#222 review: the record's *_enforced must reflect reality, not be
     hardcoded — otherwise the I7 gate can't catch a missing net boundary."""
 
+    # landlock_plan() short-circuits to None off Linux, so these three would
+    # fail with a TypeError on a macOS dev host (#242). The plan itself is
+    # pure dict arithmetic — no syscalls — so pinning the platform exercises
+    # the real semantics from any host. Keep the patch on these three only:
+    # probe_landlock_abi() guards a genuine ctypes syscall on the same flag,
+    # and forcing it "linux" off-Linux would dial an arbitrary syscall number.
+
+    @mock.patch("sys.platform", "linux")
     def test_landlock_plan_keeps_bind_ports_without_proxy(self):
         plan = MOD.landlock_plan(
             {"security": {"allow_bind_ports": [3000]}},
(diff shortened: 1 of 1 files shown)


maeste (approved):
Walked through this at the project level (logical review) — the change is sound and does exactly what #242 asks.

**Verified during review** (ran the suite on Linux and via the PR's runpy wrapper for simulated macOS, isolated clone at `30b3600e`):

- Linux: base `Ran 13 tests — OK` → head `Ran 14 tests — OK` ✓
- Simulated macOS: base `Ran 13 tests — FAILED (errors=3)` (the three `TypeError: 'NoneType' object is not subscriptable`) → head `Ran 14 tests — OK` ✓ — both cells of the table in the description reproduce exactly
- `test_landlock_plan_is_none_off_linux` is load-bearing: patching out the `sys.platform != "linux"` guard in `landlock_plan()` fails exactly that test and nothing else — good addition
- The deliberate non-class-wide scoping is right: `probe_landlock_abi()` gates a real ctypes syscall on the same flag, so patching it class-wide would be a latent hazard on Darwin

**One request before merge:** could you rebase onto current `main` and resolve conflicts as needed? The branch was cut from `8ec41542` and main has since moved (#271, including the new `scripts/tests/test_expose_gpu.py`). `test_policy_record.py` itself is conflict-free.

One note for the record: the "4 pre-existing ruff findings" only appear with an explicit `--select EXE,I,C408,SIM117`; plain `ruff check` (the repo has no ruff config) reports the file clean. Not blocking.

Approving on that basis — nice, minimal, well-scoped fix.

🤖 Walkthrough performed with Claude Code.

PR #277, opened 2026-07-31
feat(sandbox): make the snapshot store addressable across projects (#23, #24)

### LINKED ISSUE #23: snapshot-list: add --all flag to show snapshots from all projects
## Description

`agent-sandbox snapshot-list` currently only shows snapshots for the project associated with the current working directory. 

It would be useful to have a `--all` flag that lists snapshots across **all** projects, regardless of the current directory.

## Expected behavior

```bash
agent-sandbox snapshot-list --all
```

Should display all snapshots from all projects, grouped or labeled by project name.

## Current behavior

Only snapshots for the current directory's project are shown.

### LINKED ISSUE #24: snapshot-prune does not work on projects
## Description

`agent-sandbox snapshot-prune` does not seem to work correctly when operating on projects. Snapshots are not pruned as expected.

## Steps to reproduce

1. Have a project with existing snapshots
2. Run `agent-sandbox snapshot-prune`
3. Observe that project snapshots are not pruned

## Expected behavior

`snapshot-prune` should correctly prune old/unused snapshots for projects.

## Current behavior

Pruning has no effect on project snapshots.

oscerd:
Closes #23. Closes #24.

## The shared root cause

These two issues look unrelated but are the same defect seen from two sides.

Project snapshots live under `snapshots/projects/<sha256(project_path)[:12]>/`. That hash is **one-way**, and nothing anywhere recorded which project a given directory came from. So the store could only ever be addressed from *inside* the originating directory:

- **#23** — `snapshot-list` could not show another project. Worse, the issue asks for output "grouped or labeled by project name", and that was not implementable as things stood: the path had been hashed away, so the best possible output was a list of bare hashes.
- **#24** — `snapshot-prune` only ever applied `max_project_snapshots` to the cwd project's hash dir. Every *other* project, and every dir orphaned by a project that was moved or deleted, grew without bound with no way to reclaim it.

Fixing the listing without fixing the hash gets you a wall of hex; fixing prune without it gives you nothing to aim at. Hence one PR.

## A note on #24 as filed

I could not reproduce the literal report ("run `snapshot-prune`, project snapshots are not pruned"). On current `main` I built a synthetic store and ran all four flag combinations from inside and outside the project; every one pruned the cwd project correctly.

What *is* reproducible is the scoping gap above, which matches the title and is a real unbounded-growth bug. That is what this PR fixes. If the reporter was hitting something else, I'm happy to reopen with their `agent-sandbox snapshot-list --all` output — which, conveniently, this PR makes possible to ask for.

## What changed

Record the project path in a `.origin` sidecar when a project snapshot is created, then build both commands on it.

```
$ agent-sandbox snapshot-list --all --project-only
Project snapshots (all projects):

  /home/u/alpha *
    hash: f0b4f623dc47
    20260731T094810  14 B

  /home/u/beta
    hash: 29cf69c51318
    20260731T094810  13 B

  /home/u/deleted  (missing — project moved/deleted)
    hash: 0985dd0c546f
    20260731T094810  13 B

  [deadbeef1234] (unknown project — no origin recorded)
    hash: deadbeef1234
    20260101T120000  7 B

  total: 47 B
```

```
$ agent-sandbox snapshot-prune --all-projects --project-only
  /home/u/beta: removed 4
  /home/u/deleted  (missing — project moved/deleted): removed 4
  [deadbeef1234] (unknown project — no origin recorded): removed 4
Project snapshots (all 4 project(s)): removed 12, max=3 each
```

### Design decisions worth reviewing

**`--all-projects`, not `--all`, on prune.** `snapshot-prune --all` already exists and means "both project *and* config types". Repurposing it would silently change what an existing command does. The two flags are orthogonal axes and the docs now say so explicitly: `--all` picks *which kinds*, `--all-projects` picks *whose*. `snapshot-list` has no such clash, so it takes `--all` exactly as the issue requests (with `--all-projects` as an alias so the cross-cutting name works on both).

**Orphans are labelled, not hidden.** A directory whose project no longer exists is precisely the thing you want to reclaim, so it is surfaced as `(missing — project moved/deleted)` rather than filtered out.

**Pre-existing dirs keep working.** Snapshots taken before this change have no sidecar; they list and prune normally as `(unknown project — no origin recorded)`, and self-heal the next time that project is snapshotted. No migration step.

**The sidecar cannot break snapshots.** It is a file, so `list_snapshots()`'s "timestamp-shaped directory" filter ignores it (covered by a test). A failed write is swallowed — a missing *label* must never fail the *snapshot*.

## How to test

```bash
python3 scripts/tests/test_snapshot_store.py   # Ran 12 tests — OK
```

New tests cover the sidecar round-trip, unlabelled dirs, the sidecar not being mistaken for a snapshot, an unwritable store, prune reaching every project including orphans, prune idempotency, and prune keeping the newest.

Manual end-to-end, three projects with one deliberately deleted plus one legacy unlabelled dir:

```bash
for d in alpha beta gone; do agent-sandbox snapshot -p /tmp/multi/$d --project-only; done
rm -rf /tmp/multi/gone
agent-sandbox snapshot-list --all --project-only
agent-sandbox snapshot-prune --all-projects --project-only   # idempotent on re-run
```

## Backward compatibility

Default output and every existing flag are unchanged — verified across `snapshot-list`, `snapshot-list --project-only|--config-only`, `snapshot-prune`, `snapshot-prune --all|--project-only|--config-only`, plus the mutually-exclusive group still rejecting `--project-only --config-only`.

Full suite green (11/11 test files). `ruff check` on `sandbox/agent-sandbox` reports 76 findings before and after — no new ones.

### DIFF
--- a/docs/documentation/sandbox/cli-reference.md
+++ b/docs/documentation/sandbox/cli-reference.md
@@ -127,9 +127,12 @@ Create a filesystem snapshot of the project directory and/or the agent config di
 
 ```
 agent-sandbox snapshot-list [-a AGENT] [-p DIR] [--config-only | --project-only]
+                            [--all]
 ```
 
 List available snapshots grouped by type (project/config) with timestamps and disk usage.
+
+By default only the current directory's project is shown. `--all` lists every project and agent in the store, labelled by project path, with a running disk-usage total — useful when you work across several repositories and want to see where the space went.
 
 | Flag | Default | Description |
 |------|---------|-------------|
@@ -137,6 +140,28 @@ List available snapshots grouped by type (project/config) with timestamps and di
 | `-p`, `--project` | `.` (cwd) | Project directory |
 | `--config-only` | off | Show only config snapshots |
 | `--project-only` | off | Show only project snapshots |
+| `--all`, `--all-projects` | off | List every project and agent, not just the current directory |
+
+```
+$ agent-sandbox snapshot-list --all --project-only
+Project snapshots (all projects):
+
+  /home/u/alpha *
+    hash: f0b4f623dc47
+    20260731T094810  14 B
+
+  /home/u/beta
+    hash: 29cf69c51318
+    20260731T094810  13 B
+
+  /home/u/deleted  (missing — project moved/deleted)
+    hash: 0985dd0c546f
+    20260731T094810  13 B
+
+  total: 40 B
+```
+
+`*` marks the current directory's project. Snapshot directories are namespaced by a one-way hash of the project path, so each one also records the path it came from; directories created before that record existed are listed by hash as `(unknown project — no origin recorded)` and remain fully listable and prunable.
 
 ---
 
@@ -183,9 +208,12 @@ Interactive restore from a snapshot. Presents each changed file for per-file acc
 ```
 agent-sandbox snapshot-prune [-a AGENT] [-p DIR]
                              [--config-only] [--project-only] [--all]
+                             [--all-projects]
 ```
 
 Remove old snapshots beyond the configured maximum count (`max_project_snapshots`, `max_config_snapshots`).
+
+By default pruning is scoped to the current directory's project, so snapshots belonging to other projects — and to projects you have since moved or deleted — are never reclaimed. `--all-projects` prunes every project and agent in the store, applying the same per-project maximum to each.
 
 | Flag | Default | Description |
 |------|---------|-------------|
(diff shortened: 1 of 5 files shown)


maeste (approved):
Approving — the `.origin` sidecar design is the right fix for the shared root cause of #23/#24, the `--all` vs `--all-projects` axis split is well argued, and the new test suite is solid (12/12 here). Two non-blocking notes from verification:

1. **`read_snapshot_origin` (sandbox/agent-sandbox:5799)** catches only `OSError`, but `UnicodeDecodeError` is a `ValueError`: a `.origin` with any non-UTF-8 byte (torn write, disk corruption) crashes `snapshot-list --all` and `snapshot-prune --all-projects` on every run until the sidecar is removed by hand. One-line fix: `except (OSError, ValueError)` or `read_text(errors="replace")` — keeps the "a bad label must never fail the snapshot" contract you built the write side on.
2. **`cli-reference.md` prune example (~line 230)** is internally inconsistent: two per-project `removed 4` lines (8 total) against a summary of `all 4 project(s): removed 12` — the `[hash] (unknown project)` entry from the PR description got dropped from the doc examples without recomputing the counts.

Neither blocks merge; happy to take both as a fast follow-up.

Issue #222, opened 2026-06-10
Integrate Landlock enforcement into agent-sandbox (__landlock-exec) per the #211 spike GO


maeste:
Follow-up implementation of the Landlock spike (#211 / PR #218), whose verdict is **GO**. Framing unchanged from the spike: Landlock is a **hardening layer for the existing bwrap backend, not a second sandbox product** — bwrap paints the mount picture, Landlock adds a kernel fence inside it.

## Scope
Fold the proven prototype (`sandbox/spikes/landlock/landlock_exec.py`) into `sandbox/agent-sandbox`:
- `__landlock-exec` internal subcommand, injected as the last bwrap argv element (the spike experimentally rejected `preexec_fn`: bwrap-created tmpfs is not beneath host-opened rule FDs); opens rule FDs in the final mount namespace, applies fs (+ TCP net on ABI ≥ 4) ruleset, execs the agent. Graceful ABI probing.
- Default enablement per the spike recommendation: **paranoid** = fs + net (connect → proxy bridge 8118 only); **normal** = fs + net-when-proxy-active; **permissive** = off.
- **Fail-closed at paranoid** (invariant I7 of the Config v2 design doc): if the requested fs boundary cannot be enforced (no Landlock), the launch fails naming the missing boundary; degraded operation only as explicit opt-in. Emits the effective-policy record fields it owns (fs/net enforced, ABI, degraded_reason) — full record assembly is #221.
- Config keys: align with the `[filesystem]`/`[network]` policy keys defined in the Config v2 design doc (#201, §2/§4.4 golden test: same intent → bwrap mounts AND Landlock rules). Pre-v2 interim keying on sandbox levels is acceptable if this lands before the policy layer.
- `sandbox/tests/` check asserting enforcement (the spike's gate assertions, ported).

## Validation gates (from the spike's validation plan)
- Re-run probe + demo + paranoid gate on **Ubuntu 24.04 GA (kernel 6.8, ABI v4 — the distro floor)**: validates TCP net rules at exactly the minimum ABI. The spike proved ABI v7 (Fedora 44) live; v4 is cited, not yet probed.
- No-Landlock kernel path: degraded record + fail-closed behavior (simulated via `LINCE_LANDLOCK_FORCE_ABI=0` in the spike; a real `lsm=`-without-landlock boot or container is the stronger check).

## Relations
[...]

PR #290, opened 2026-08-24
fix(sandbox): use sys.executable for Landlock shim relaunch

siculo:
## Summary

`agent-sandbox run` can fail with `Error: Python 3.11+ required (for tomllib)` even when the outer Python interpreter is 3.11+.

The Landlock hardening shim (#222) relaunches `agent-sandbox` a second time *inside* the bwrap sandbox, using the hardcoded interpreter name `"python3"` (`build_bwrap_cmd()`, `sandbox/agent-sandbox`):

```python
shim: list[str] = ["python3", LANDLOCK_SHIM_MOUNT, "__landlock-exec"]
```

This name is resolved through the sandbox's own `PATH`, which is rebuilt from scratch and puts system directories (`/usr/local/bin`, `/usr/bin`, `/bin`) before any user/conda directories. So on a system where the system `python3` is older than 3.11 (for example Ubuntu 22.04) but the user is running the outer process from a newer interpreter (e.g. a conda env), the inner relaunch picks the wrong, older `python3` and fails on `import tomllib`.

Since the whole host filesystem is read-only bind-mounted at the same path inside the sandbox (`--ro-bind / /`), an absolute interpreter path stays valid there. This PR replaces the hardcoded `"python3"` with `sys.executable`, so the shim always reuses the exact interpreter that is already running the outer process, independent of what the sandbox's `PATH` exposes.

## Change

```diff
-        shim: list[str] = ["python3", LANDLOCK_SHIM_MOUNT, "__landlock-exec"]
+        shim: list[str] = [sys.executable, LANDLOCK_SHIM_MOUNT, "__landlock-exec"]
```

`sys` is already imported at the top of the script, so no new import is needed.

## How to test

1. Use an interpreter for the outer process that is not the system `python3` but is 3.11+ (e.g. a conda env), on a system whose system `python3` is older than 3.11.
2. Run `agent-sandbox run` with a config that enables the Landlock shim (`__landlock-exec`, feature #222).
3. Before the fix: the inner relaunch fails with `Error: Python 3.11+ required (for tomllib)`.
4. After the fix: the inner relaunch succeeds, using the same interpreter as the outer process.

Also verified `python3 sandbox/agent-sandbox --help` still works after the change.

## Backward compatibility

No behavior change for users whose outer interpreter is already `/usr/bin/python3` (or otherwise the same one the sandbox `PATH` would have resolved to). No new dependencies, no changes to the CLI surface.

### DIFF
--- a/sandbox/agent-sandbox
+++ b/sandbox/agent-sandbox
@@ -3400,7 +3400,7 @@ def build_bwrap_cmd(config: dict, project_dir: Path,
             cmd[-1:-1] = opts
         else:
             cmd += opts
-        shim: list[str] = ["python3", LANDLOCK_SHIM_MOUNT, "__landlock-exec"]
+        shim: list[str] = [sys.executable, LANDLOCK_SHIM_MOUNT, "__landlock-exec"]
         for target in bwrap_rw_targets(cmd):
             shim += ["--rw", target]
         shim += ["--ro", "/"]


sandbox/agent-sandbox
```diff
@@ -3400,7 +3400,7 @@ def build_bwrap_cmd(config: dict, project_dir: Path,
             cmd[-1:-1] = opts
         else:
             cmd += opts
-        shim: list[str] = ["python3", LANDLOCK_SHIM_MOUNT, "__landlock-exec"]
+        shim: list[str] = [sys.executable, LANDLOCK_SHIM_MOUNT, "__landlock-exec"]
```
maeste, line 3403: The direction is right for the reported bug, but the reasoning in the PR description doesn't hold for `$HOME`, and there is a regression class worth handling: the absolute interpreter path is only guaranteed to survive inside the sandbox for paths outside `$HOME`. In-namespace, the mounts are applied in this order: `--ro-bind / /`, then a fresh tmpfs over the whole home (`--tmpfs <home>`), then selective read-only re-binds of the top-level `$HOME` dirs found in `$PATH` (`detect_home_path_dirs()`), gated on `auto_expose_path`. So `sys.executable` resolves in-namespace only when the interpreter's prefix is outside `$HOME`, or under a PATH-derived top dir — which is why an *activated* conda/venv works here (by coincidence of those PATH-derived binds, not because of `--ro-bind / /`).

Concretely, users this can now break — all of whom worked before when system python3 ≥ 3.11, since `python3` resolved to `/usr/bin/python3` via the sandbox PATH:
1. `auto_expose_path = false` + any `$HOME` interpreter (conda's default install location is `~/miniconda3`);
2. a venv created from a conda base — `sys.executable` is the venv python, whose symlink target in the (unbound) conda base dangles in the namespace → bwrap exec fails with a bare "No such file or directory";
3. explicit-path invocation with the env not on PATH (`~/someenv/bin/python agent-sandbox run`).

In all three the agent never starts, and the shim's own error path can't help because it never runs. Suggested follow-up: fall back to `"python3"` when `sys.executable` is under `$HOME` and not under an exposed top dir, and adjust the PR description's mount reasoning accordingly. Not blocking — approving, since the primary target population (activated conda on hosts with old system python) is strictly better off with this change.

maeste (approved):
Reviewed: the fix is correct for the reported bug (the sandbox PATH puts system dirs first, so the hardcoded `python3` could resolve to an interpreter older than 3.11), and no test pins the argv literal. One non-blocking inline comment about the mount reasoning and a regression class for `$HOME`-based interpreters outside the PATH-derived binds — worth a follow-up.

PR #270, opened 2026-06-25
feat(lince-lab): macOS-guest backend via Tart (#263)

rinaldofesta:
Closes #263. First brick of the lince-lab v2 (macOS/Seatbelt) epic, #268. macOS host / Apple Silicon only.

## What this adds

A `MacosBackend` that boots a clean disposable macOS guest via Tart (Virtualization.framework) behind the existing broker verb contract, selected with `LINCE_LAB_MACOS=1`. It mirrors `lima_backend.py`; broker/policy/capture-driver/CLI/protocol are unchanged, and recipe/bisect change by one line each (the egress seam below).

Per-verb mapping (`lince_lab/macos_backend.py`):
- create: `tart clone <oci-ref> <name>` then `tart set --cpu --memory` (caps floored to cpu >= 2, mem >= 4096 MB: the Linux-tuned 2 GiB default does not boot macOS reliably, and `tart set --memory` is absolute so it would downgrade the image). A non-OCI location (the Linux default_image qcow2 URL) is rejected with a `pass --image macos-sequoia` message rather than a cryptic `tart clone` error.
- start: `tart run --no-graphics` is a foreground process, so it is spawned detached, its output captured to a log, and reaped on stop/delete. It then waits for the DHCP lease (`tart ip`) and for sshd to answer before returning, and fails fast if the run process exits early instead of hanging the wait window.
- exec/copy: over SSH/scp to `tart ip` (guest admin password via `SSH_ASKPASS`, no sshpass dependency). exec returns the guest exit code verbatim (the bisect signal).
- snapshots: Virtualization.framework has no loadvm, so they are emulated with APFS copy-on-write `tart clone` of a stopped golden guest (`<vm>__snap__<tag>`). This keeps the bisect reset contract: one `snapshot_apply(BASE)` per candidate.
- capture: `ht` in the guest over SSH, reusing the existing stdio-pipe channel.

Supporting changes:
- `config.py` gains the `macos-sequoia` image (Tart OCI ref `ghcr.io/cirruslabs/macos-sequoia-base:latest`, arch arm64).
[...]

Issue #266, opened 2026-06-15
lince-lab v2: Seatbelt test recipes


maeste:
## Context

**Seatbelt test recipes** — the macOS equivalents of the v1 Lince fixtures, asserting the Seatbelt sandbox behaves correctly in a clean macOS guest. Part of the **lince-lab v2** epic.

## Trigger / Entry condition

- **Depends on:** #265, #259, #256 (must be closed).
- **Machine gate:** `bash scripts/lince-lab/ci/v2/03-contract.sh` && `bash scripts/lince-lab/ci/08-fixtures.sh` && `bash scripts/lince-lab/ci/05-capture.sh` all exit `0`.
- **Host gate:** real run on Darwin/Apple Silicon; skips with exit `0` elsewhere.

## Scope

- Seatbelt recipes mirroring v1: clean install on macOS guest, dashboard wizard agent-list assertion, installer idempotency — adapted to the Seatbelt path.

## Acceptance criteria

- [ ] Seatbelt `wizard-n-agent-list` and `installer-idempotency` recipes pass on a clean macOS guest.

## Validation — automated, no human

- `bash scripts/lince-lab/ci/v2/04-recipes.sh` → on Darwin: run the Seatbelt recipes → exit `0`. On non-Darwin: skip, exit `0`.

## Definition of Done

- [ ] Validation green on macOS; skips on Linux.
- [ ] Recipes committed under `lince-lab/recipes/macos/`.
- [ ] PR opened with `Closes #<this issue>`; repo CI green.

## Exit condition (loop stop signal)

- `bash scripts/lince-lab/ci/v2/04-recipes.sh` exits `0`. Trigger oracle for #267.

PR #304, opened 2026-09-02
fix(lince-lab): address macOS backend review follow-ups

rinaldofesta:
Follow-up to #270.

`MacosBackend.exec()` returned OpenSSH exit 255 for both transport failures and a real guest exit 255. In a bisect, a dropped SSH connection could therefore become a false bad verdict.

This patch:

- wraps the remote command and returns its status in a nonce-scoped stderr frame; guest 255 remains 255, while a missing frame or nonzero SSH status raises `BackendError`
- removes every `<vm>__snap__*` Tart clone before deleting its parent VM
- states the temporary macOS fail-open exception in `apply_egress_lockdown()` until #266

I did not use retry-on-255: recipe steps can mutate state, so running one twice is unsafe.

Validation:

- `python3 -m unittest discover -s lince-lab/tests -p 'test_*.py'` -> 262 passed, 16 skipped
- `python3 lince-lab/tests/test_macos_backend.py` -> 26 passed
- `python3 -m compileall -q lince-lab/lince_lab lince-lab/tests`
- `git diff --check`

Not run: `ruff`; it is not installed in this environment.

### DIFF
--- a/lince-lab/lince_lab/macos_backend.py
+++ b/lince-lab/lince_lab/macos_backend.py
@@ -28,6 +28,7 @@ from __future__ import annotations
 import atexit
 import json
 import os
+import secrets
 import shlex
 import subprocess
 import sys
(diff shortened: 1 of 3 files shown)


maeste (approved):
Both follow-ups land as described — the nonce-scoped stderr frame is the right shape for the 255 disambiguation (and skipping retry-on-255 is the correct call for state-mutating steps), the golden cascade closes the orphaned-disk gap, and the docstring states the deferral. Tests verified here (262/16) and ruff is clean on all three files. Two non-blocking notes from verification:

1. **Frame parser requires the marker to be the last stderr bytes** (`macos_backend.py:464`): a backgrounded child that keeps the ssh stderr pipe open and writes after the frame (e.g. `nohup ./server &; make test`) turns a *successful* exec into `BackendError: invalid ssh guest-exit frame` — flaky, since it races the frame. Tolerating trailing garbage after the frame's terminating newline keeps transport detection intact.
2. **The snapshot cascade in `delete()` is fail-hard** (`macos_backend.py:259`): `snapshot_list`/`snapshot_delete` raise while `stop()` two lines up is best-effort, so one non-deletable clone (e.g. a golden someone `tart run`s manually) permanently blocks deleting the parent — and `vm list` hides `__snap__` names, so the CLI offers no way to see or stop the offending clone. Best-effort per-clone deletes (warn, then delete the parent) would match the surrounding pattern.

Taking both as fast follow-ups; merging.

